PR #16, opened 2026-01-17
Update GitHub Actions workflow for Go project

SuperCoolPencil:
Add github actions to
- test: Runs on normal pushes and PRs to check code.
- release: Runs when a tag is pushed, utilizing GoReleaser and Homebrew token.


SuperCoolPencil:
- update .goreleaser.yaml to include homebrew config
- update github actions to automatically run goreleaser when a new tag is pushed
- update readme to add install method homebrew

PR #18, opened 2026-01-18
Update Checker

SuperCoolPencil:
feat: implement automatic version update checking with TUI notification and user settings.


SuperCoolPencil:
image
image

PR #19, opened 2026-01-18
test: Added tests for settings.go

junaid2005p:
Added more tests for settings.go improving code coverage for config folder to 65.2% from 50%.
Removed OS specific filepath tests which would fail based on different OS


SuperCoolPencil (approved):
lgtm

PR #20, opened 2026-01-18
test: Added tests for pool.go

junaid2005p:
pool.go did not have a test file. Added pool_test.go with tests for pool.go and improved test coverage for download folder.


SuperCoolPencil (commented):
lgtm

PR #29, opened 2026-01-19
Intelligent 429

SuperCoolPencil:
implement rate limiting for concurrent downloads, handling 429 responses with Retry-After and exponential backoff.

### DIFF
--- a/internal/download/manager.go
+++ b/internal/download/manager.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


SuperCoolPencil:
image

SuperCoolPencil:
letting workers hammer the server with requests seems to be better for performance than "intelligent"ly retrying. maybe this was just bad implementation. 👎🏻

SuperCoolPencil:
test with github actions?

PR #30, opened 2026-01-20
feat: Add `--compare-global` option to benchmark globally installed surge against local build for better testing

SuperCoolPencil:


SuperCoolPencil:
basically if you pass --compare-global flag it will compare the global version of surge (ie. previous release) with ./surge (ie. new build) 

extremely helpful to compare performance between 2 versions!

PR #38, opened 2026-01-25
fix: metadata zip error

### LINKED ISSUE #23: Sourceforge download metadata error.
Sourceforge downloads are grabbing name as `metadata.zip`.

Image

elijahthis:
Fixes #23  
### Summary
This PR fixes an issue where downloads were being named after the internal contents of a ZIP file instead of the intended external filename.

### Problem
In `filename.go`, the `DetermineFilename` function was occasionally prioritizing the ZIP Local Header over more relevant sources. For example, a download from https://sourceforge.net/projects/crdroid/files/alioth/12.x/crDroidAndroid-16.0-20260114-alioth-v12.5.zip/download was being saved as `metadata.zip` because `metadata` was the first entry inside the archive.

### Fix
Reordered the filename discovery logic to prioritize External Metadata over Internal Content. The new priority is:
1. Content-Disposition
2. Query Parameters
3. URL Path
4. ZIP Local Header

image

### DIFF
--- a/internal/utils/filename_test.go
+++ b/internal/utils/filename_test.go
@@ -1,6 +1,9 @@
 package utils
 
 import (
+	"bytes"
+	"io"
+	"net/http"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


junaid2005p:
Hi thank you for the PR. Yes this is an issue. The fix looks good to me.
However I was thinking maybe we should add a test to filename_test.go to ensure this does not happen again

Something like

```
func TestDetermineFilename_ContentDispositionOverwrittenByZip(t *testing.T) {
	internalFilename := "internal/path/wrong_name.txt"
	zipHeader := make([]byte, 30+len(internalFilename))

	// Signature: PK\x03\x04
	zipHeader[0] = 0x50
	zipHeader[1] = 0x4B
	zipHeader[2] = 0x03
	zipHeader[3] = 0x04

	// Filename length at offset 26-27 (little-endian)
	zipHeader[26] = byte(len(internalFilename))
	zipHeader[27] = 0

	// Extra field length at offset 28-29 (0)
	zipHeader[28] = 0
	zipHeader[29] = 0

	// Filename starts at offset 30
	copy(zipHeader[30:], internalFilename)

	// Create a mock HTTP response with Content-Disposition header
	resp := &http.Response{
		Header: http.Header{
			"Content-Disposition": []string{`attachment; filename="my_archive.zip"`},
		},
		Body: io.NopCloser(bytes.NewReader(zipHeader)),
	}

	filename, _, err := DetermineFilename("https://example.com/download", resp, false)
	if err != nil {
		t.Fatalf("DetermineFilename failed: %v", err)
	}

	// Currently returns "wrong_name.txt" (from ZIP internal path)
	// Should return "my_archive.zip" (from Content-Disposition)
	expected := "my_archive.zip"
	if filename != expected {
		t.Errorf(
			"DetermineFilename() = %q, want %q (Content-Disposition should take precedence over ZIP internal name)",
			filename,
			expected,
		)
	}
}
```

elijahthis:
Hii @junaid2005p. Just added tests

PR #41, opened 2026-01-25
feat: Randomize benchmark task order, increase inter-task cooldown, and run more iterations in CI.

SuperCoolPencil:
surge (current) always seems to have better performance than surge (baseline) even when they refer to the same code. 

I suspect this is related to the execution order, as surge (current) was always run first so benchmarking had a order bias. 

This PR tries to eliminate the bias by introducing randomness. Also increased cooldown between downloads to prevent disk caching and allow the token bucket to refill.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -33,6 +33,7 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "1.24.4+auto"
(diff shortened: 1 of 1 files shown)


SuperCoolPencil:
well, its not completely fixed but it's better than before...

junaid2005p (approved):
Lgtm

Issue #57, opened 2026-01-29
[Bug] Extension triggers re-download of entire browser download history upon installation


yi427:
**Describe the bug**
I installed the Surge browser extension (Chrome). When the server activation, the extension intercepted and started re-downloading ALL files from my browser's past download history, not just the active ones.

**To Reproduce**
1. Have a browser history with multiple finished downloads.
2. Start the surge server
3. the queue occur files form my google install history 

**Expected behavior**
When I install new file, the surger to download.

**Screenshots**
Sorry, I delete all file, so I don't to provite the screenshots.

**Please complete the following information:**
 - OS: [MacOs 26.3Beta]
 - Installed From: [Brew]

PR #59, opened 2026-01-29
fix: Extension retriggers old downloads

SuperCoolPencil:
### This PR addresses two critical issues:
- History Re-download: The extension was intercepting ALL past downloads from the browser history upon startup/activation. #57 
- Filename Duplication: Downloads were occasionally receiving malformed names like file(3)(1).bin instead of correctly incrementing to file(4).bin.

### Solved them like this: 

--
- Problem: chrome.downloads.onCreated fires for historical items (e.g., when the extension loads)
- Fix: Added checks to ignore download unless state is in_progress and if started > 60s ago
--
- Problem: The backend's uniqueFilePath function in manager.go failed to parse existing file counters (e.g., file (3).pdf) if they contained trailing whitespace or formatting irregularities common with browser downloads. This caused it to fallback to appending a new counter (1), resulting in file (3)(1).pdf.
- Fix: Updated manager.go to robustly TrimSpace and parse the suffix.


junaid2005p (approved):
LGTM

PR #81, opened 2026-02-04
feat: change engine to use range assignment instead of dynamic chunking

SuperCoolPencil:
This PR refactors the Surge engine to switch from a Dynamic Chunking model (many small tasks per worker) to a Range Assignment model (one large task per worker, with work stealing).

The engine now assigns one large task per worker (fileSize / numWorkers). This maximizes sequential throughput while relying on the existing work-stealing logic to balance the load if speeds are uneven.


SuperCoolPencil:
This PR also adds a new feature for sequential downloads, now you can watch media files as they are being downloaded

PR #154, opened 2026-02-11
Settings bug

junaid2005p:
Fixed these settings issues

1. Max Global Connections was shown as a setting in the TUI. However it was not functional. All instances of it were removed.
2. ProxyUrl functionality was there but actually nothing was happening since it wasnt being copied into types.RuntimeConfig being used by the Downloader
3. Cleaned some commented code in root.go


SuperCoolPencil (approved):
nice

PR #163, opened 2026-02-12
test: Added more tests for scenarios not being covered

junaid2005p:


junaid2005p:
I used a workerpool in local_service.go to simulate goroutines needed for tests in local_service_test.go

However this was not the right way. So closing this for now.

PR #172, opened 2026-02-13
fix: Pause downloads if terminal crashes or is abruptly exited

### LINKED ISSUE #167: When terminal is closed without pausing and exiting with ctrl + Q, The download just dissapears and doesnt appear in the list
**Describe the bug**
When Terminal is closed without pausing and exiting with ctrl + Q, The download just dissapears and doesnt appear in the list

**To Reproduce**
Start a download and just close the terminal

**Expected behavior**
The download should atleast pause 

 - OS: Arch Linux
 - Version 0.5.4-1
 - Installed From: AUR

junaid2005p:
Fixes #167

Added a new cmd/shutdown.go to contain graceful shutdown handling.
both the TUI and the server now open a buffered channel and listen to os.SIgnals

if any exit happens they call shutdown.go to handle it

Also local_service.go exit handling is now idempotent so multiple exit calls will not be an issue

### DIFF
--- a/cmd/shutdown_test.go
+++ b/cmd/shutdown_test.go
@@ -1,0 +1,32 @@
+package cmd
+
+import (
+	"sync"
+	"sync/atomic"
+	"testing"
+)
+
+func TestExecuteGlobalShutdown_Once(t *testing.T) {
+	var calls int32
+	resetGlobalShutdownCoordinatorForTest(func() error {
+		atomic.AddInt32(&calls, 1)
+		return nil
+	})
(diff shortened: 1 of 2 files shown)


SuperCoolPencil:
Should we handle all signals in the same way? 

I think SIGKILL, SIGSEGV should be handled differently, no?

SuperCoolPencil:
> Should we handle all signals in the same way? 
> 
> I think SIGKILL, SIGSEGV should be handled differently, no? 

nvm I was wrong, these flags cannot be handled by an app.

junaid2005p:
We cannot handle SIGKILL reliably as the OS closes the process immediately

PR #189, opened 2026-02-15
fix: adds auth for cli commands when communicating with surge server

### LINKED ISSUE #186: Server error: unauthorised to add download
**Describe the bug**
When running a Surge server locally, I can access it but not add downloads to it if I do not go into the TUI.
My token is set into `SURGE_TOKEN`..

I tried `surge add https://nbg1-speed.hetzner.com/1GB.bin` and got:
```bash
Error adding https://nbg1-speed.hetzner.com/1GB.bin: server error: 401 Unauthorized - Unauthorized
```

My goal was to use `surge` instead of `aria2c` inside my script that uses Real-Debrid for torrent.
I would have loved to run the TUI in one terminal and run my script in another to send it downloads to perform.

I can connect from another terminal via `surge connect` without issue, and when a download is started the TUI shows the correct `token`.

**To Reproduce**
Steps to reproduce the behaviour:
1. Start surge server in one terminal with `surge server start`.
2. Open another terminal and check server status with `surge server status`. It should output `Surge server is running (PID: 158306, Port: 1700).`
3. Try adding download to list with `surge add https://nbg1-speed.hetzner.com/1GB.bin`
4. Get error `Error adding https://nbg1-speed.hetzner.com/1GB.bin: server error: 401 Unauthorized - Unauthorized`

**Expected behaviour**
I thought we could send download links to the Surge server from another terminal but it seems I can only add downloads via the TUI.

 - *Version*:  0.6.7
 - *OS*: Kubuntu 25.10 (Questing Quokka) x86_64
 - *Kernel*: Linux 6.17.0-12-generic
 - *Terminal*: konsole 25.8.1
 - *Installed From*: Github release file `surge_0.6.7_linux_amd64.tar.gz`

junaid2005p:
Fixes #186 

We now have a new doLocalAPIRequest() which adds the auth token. All CLI commands now use this function

### DIFF
--- a/cmd/cli_test.go
+++ b/cmd/cli_test.go
@@ -104,6 +104,80 @@ func TestResolveDownloadID_RemoteStillWorksWhenDBUnavailable(t *testing.T) {
 	}
 }
 
+func TestResolveDownloadID_StrictRemoteDoesNotFallbackToDBOnRemoteError(t *testing.T) {
+	setupIsolatedCmdState(t)
+
+	entry := types.DownloadEntry{
+		ID:       "11223344-1234-5678-90ab-cdef12345678",
+		Filename: "db-only.bin",
+	}
+	if err := state.AddToMasterList(entry); err != nil {
+		t.Fatalf("failed to seed db entry: %v", err)
+	}
+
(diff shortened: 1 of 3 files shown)


SuperCoolPencil (approved):
done

PR #191, opened 2026-02-15
Harden Dockerfile: pin Alpine version, fix ARM arch, add OCI labels and config dirs

elysiummachines:
## Changes
- Pin base image from `alpine:latest` to `alpine:3.21` for reproducible builds
- Fix `armv7l` architecture mapping from `arm64` to `arm` (correct 32-bit ARM target)
- Add OCI image labels for better maintainability and discoverability
- Update `mkdir` to include `/root/.local/state/surge/logs` where surge actually stores runtime data

## Why
- `alpine:latest` is unpredictable and can cause unexpected breakage on new releases ( I've been there to many times)
- `armv7l` is 32-bit ARM and should map to `arm` not `arm64`
- OCI labels follow industry standard for container image metadata
- Proper config dirs prevent permission errors on container startup

## Testing
- Built and tested locally on `amd64`
- Container starts successfully
- Token accessible at `docker exec surge cat /root/.local/state/surge/token` (this works across all paths includes in YAML)
- `surge server status` healthcheck passes

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:3.21
 
 RUN apk add --no-cache ca-certificates curl jq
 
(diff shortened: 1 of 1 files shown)


elysiummachines:
I can take responsibilities for testing new alpine versions and updating the docker file once a full test has passed

SuperCoolPencil (approved):

SuperCoolPencil:
Thank you so much for contributing!

elysiummachines:
np at all

PR #244, opened 2026-03-18
feat: Add download history and file-opening features

FHYQ-Dong:
## 1. Short summary
This PR improves extension usability and backend safety for file-opening actions. It adds explicit `Active/History` views in both browser extensions, introduces guarded open-action APIs on the backend, and returns `/history` in recent-first order. It also wires the new remote-open control into settings and TUI.

## 2. Backend CLI/API changes
- In `cmd/http_api.go`, `/history` is now sorted by newest completion first (`CompletedAt` desc, with `ID` tie-breaker).
- Added two endpoints:
  - `POST /open-file?id=<downloadId>`
  - `POST /open-folder?id=<downloadId>`
- Added `resolveDownloadDestPath(service, id)` to resolve destination path from `GetStatus` first, then `History` fallback.
- Added `ensureOpenActionRequestAllowed(r)`:
  - loopback requests are allowed by default;
  - non-loopback requests require `general.allow_remote_open_actions=true`.
- Added new config field `general.allow_remote_open_actions` (default `false`) and wired it in:
  - `internal/config/settings.go`
  - `internal/tui/settings_view.go`
  - `docs/SETTINGS.md`

## 3. Extension changes
- Chrome and Firefox popup now have explicit `Active` / `History` tabs:
  - `extension-chrome/popup.{html,css,js}`
  - `extension-firefox/popup.{html,css,js}`
- Popup state/model split into `activeDownloads`, `historyDownloads`, and `currentView`.
- Added view-specific rendering helpers and empty-state behavior.
- Background scripts now support:
  - `getHistory` (`GET /history`)
  - `openFile` (`POST /open-file`)
  - `openFolder` (`POST /open-folder`)
- Popup action buttons now include `open-folder` and `open-file` for completed items.

## 4. Added test files
- Added: `cmd/http_api_test.go`
  - `TestEnsureOpenActionRequestAllowed_RemoteToggle`
  - `TestHistoryEndpoint_SortsMostRecentFirst`

Also updated existing tests:
- `internal/config/settings_test.go` (asserts `AllowRemoteOpenActions` defaults to `false`)

<h3>Greptile Summary</h3>

This PR extends Surge with download history viewing and remote file/folder opening. It adds `/open-file` and `/open-folder` API endpoints guarded by a loopback check and a new `allow_remote_open_actions` setting, sorts `/history` newest-first, and gives both browser extensions explicit `Active`/`History` tab views with open-action buttons for completed downloads.

**Key changes:**
- `cmd/http_api.go`: New `/open-file` and `/open-folder` POST endpoints with `ensureOpenActionRequestAllowed` (rejects non-loopback unless opt-in) and `resolveDownloadDestPath` (falls back from `GetStatus` to history). Three sentinel errors (`ErrServiceUnavailable`, `ErrDownloadNotFound`, `ErrNoDestinationPath`) give clean error mapping — though `ErrNoDestinationPath` currently falls through to `500 Internal Server Error` rather than a semantically appropriate `422`.
- `internal/utils/open.go`: New platform-aware opener correctly uses `open` (macOS), `cmd /c start` (Windows), and `xdg-open` (Linux).
- Extension popups (Chrome + Firefox): History tab is lazily loaded and rate-limited to 15 s; active poll correctly scopes to the current view. A logic bug causes the `> 1 week` ETA label to be unreachable — any duration over 1 day is always shown as `"> 1 day"` because the day guard fires first.
- Tests: Comprehensive for `resolveDownloadDestPath` edge cases and status-code mapping; `buildOpenCommand`'s per-platform routing is untested.

- Safe to merge with minor fixes — one logic bug causes incorrect ETA labels, and one sentinel error maps to a misleading HTTP status code.
- The backend changes are well-structured and correctly secured. The main issues are: (1) an unreachable `> 1 week` branch in `formatETA` (both extensions) that degrades UI readability for long ETAs, and (2) `ErrNoDestinationPath` returning 500 instead of a more accurate status. Neither is a security issue or data-loss risk, but both are real functional defects introduced in this PR.
- extension-chrome/popup.js and extension-firefox/popup.js (formatETA logic bug); cmd/http_api.go (ErrNoDestinationPath status code mapping)

| Filename | Overview |
|----------|----------|
| cmd/http_api.go | Adds `/open-file` and `/open-folder` endpoints with loopback guard, history sort, and sentinel-error mapping — well-structured, though `ErrNoDestinationPath` falls through to 500 rather than a more descriptive status code. |
| cmd/http_api_test.go | Thorough table-driven tests for `resolveDownloadDestPath`, endpoint status-code mapping, remote-toggle guard, and forwarded-loopback denial — good coverage of new functionality. |
| extension-chrome/popup.js | Adds Active/History tab switching and open-file/open-folder actions; contains an unreachable `> 1 week` branch in `formatETA` and section-labeling comments that restate the code rather than explaining intent. |
| extension-firefox/popup.js | Mirror of Chrome popup changes for Firefox (browser API instead of chrome API); carries the same unreachable `formatETA` branch and "what" comments as the Chrome version. |
| extension-chrome/background.js | Adds `fetchHistoryList`, `openDownloadFile`, and `openDownloadFolder` with correct auth headers and error propagation; comment on line 337 confirms the redundant client-side sort from the prior review was removed. |
| extension-firefox/background.js | Firefox background script equivalent of Chrome changes; uses manual `AbortController` timeout pattern consistent with Firefox's MV2 constraints. |
| internal/utils/open.go | New platform-aware file/folder opener using the correct system commands (`open`, `cmd /c start`, `xdg-open`); properly validates paths and detaches the subprocess with a goroutine to avoid blocking. |
| internal/utils/open_test.go | Tests validation guards but leaves `buildOpenCommand`'s platform-routing logic untested; adding even a smoke test for the current-platform branch would prevent regressions. |
| internal/config/settings.go | Cleanly adds `AllowRemoteOpenActions` to `GeneralSettings` with metadata, description, and a correct `false` default; no issues found. |
| internal/tui/settings_view.go | Settings TUI wires the new `allow_remote_open_actions` toggle via metadata-driven rendering; no structural changes needed. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Ext as Browser Extension
    participant BG as Background Script
    participant API as HTTP API (cmd/http_api.go)
    participant Guard as ensureOpenActionRequestAllowed
    participant Resolver as resolveDownloadDestPath
    participant Opener as utils.OpenFile / OpenContainingFolder

    Ext->>BG: openFile / openFolder (id)
    BG->>API: POST /open-file?id=... (or /open-folder)
    API->>Guard: check RemoteAddr + XFF headers
    alt loopback with no proxy headers
        Guard-->>API: nil (allowed)
    else non-loopback or proxied
        Guard->>API: check AllowRemoteOpenActions setting
        alt setting disabled
            API-->>BG: 403 Forbidden
            BG-->>Ext: { success: false, error }
        end
    end
    API->>Resolver: resolveDownloadDestPath(service, id)
    alt GetStatus returns valid DestPath
        Resolver-->>API: destPath
    else fallback to History
        Resolver->>API: scan history entries
        alt found with path
            Resolver-->>API: destPath
        else not found / no path
            Resolver-->>API: ErrDownloadNotFound / ErrNoDestinationPath
            API-->>BG: 404 / 500
        end
    end
    API->>Opener: OpenFile(destPath) or OpenContainingFolder(destPath)
    Opener-->>API: nil or error
    API-->>BG: 200 OK { status: "ok" } or 500
    BG-->>Ext: { success: true/false }
```

<details open><summary><h3>Comments Outside Diff (10)</h3></summary>

1. `extension-firefox/background.js`, line 1087-1101 ([link](https://github.com/surge-downloader/surge/blob/036aade4dd91aa71eb469e6b3fd93c9137c28705/extension-firefox/background.js#L1087-L1101)) 

   <a href="#">P2</a> **`clearTimeout` not called on fetch failure**

   If `fetch` throws (e.g., network error or the signal abort itself fires synchronously), `clearTimeout(timeoutId)` on line 1095 is never reached. The timer then fires 5 seconds later and calls `controller.abort()` on an already-settled request — harmless but leaks a live timer on every failure.

   The Chrome version avoids this entirely with `AbortSignal.timeout(5000)`. For Firefox, wrapping `clearTimeout` in a `finally` block is the idiomatic fix:

   ```js
   const controller = new AbortController();
   const timeoutId = setTimeout(() => controller.abort(), 5000);
   try {
     const headers = await authHeaders();
     const response = await fetch(`${baseUrl}/open-file?id=${encodeURIComponent(id)}`, {
       method: 'POST',
       headers,
       signal: controller.signal,
     });
     if (response.ok) return { success: true };
     const message = await response.text().catch(() => 'Failed to open file');
     return { success: false, error: message };
   } catch (error) {
     return { success: false, error: error.message };
   } finally {
     clearTimeout(timeoutId);
   }
   ```

   The same pattern applies to `openDownloadFolder` and `fetchHistoryList` in this file.

   `````markdown
   Path: extension-firefox/background.js
   Line: 1087-1101

   Comment:
   **`clearTimeout` not called on fetch failure**

   If `fetch` throws (e.g., network error or the signal abort itself fires synchronously), `clearTimeout(timeoutId)` on line 1095 is never reached. The timer then fires 5 seconds later and calls `controller.abort()` on an already-settled request — harmless but leaks a live timer on every failure.

   The Chrome version avoids this entirely with `AbortSignal.timeout(5000)`. For Firefox, wrapping `clearTimeout` in a `finally` block is the idiomatic fix:

   ```js
   const controller = new AbortController();
   const timeoutId = setTimeout(() => controller.abort(), 5000);
   try {
     const headers = await authHeaders();
     const response = await fetch(`${baseUrl}/open-file?id=${encodeURIComponent(id)}`, {
       method: 'POST',
       headers,
       signal: controller.signal,
     });
     if (response.ok) return { success: true };
     const message = await response.text().catch(() => 'Failed to open file');
     return { success: false, error: message };
   } catch (error) {
     return { success: false, error: error.message };
   } finally {
     clearTimeout(timeoutId);
   }
   ```

   The same pattern applies to `openDownloadFolder` and `fetchHistoryList` in this file.

   `````
   </details>

2. `cmd/http_api.go`, line 94-95 ([link](https://github.com/surge-downloader/surge/blob/8c3e76cf36bf575e2700697d4c7e798c7860c695/cmd/http_api.go#L94-L95)) 

   **`statusCodeForResolveDownloadError` is defined but never called — tests will fail**

   `statusCodeForResolveDownloadError` (lines 267–276) maps `ErrServiceUnavailable` → 503, `ErrDownloadNotFound` → 404, and everything else → 500. Both the `/open-file` and `/open-folder` handlers ignore it and hardcode `http.StatusInternalServerError` (500) for every error. This directly contradicts `TestOpenEndpoints_ReturnMappedResolveStatuses`, which asserts 503 for a nil service and 404 for a missing download — both tests will fail against the current implementation.

   The fix is to replace the hardcoded status in both handlers with the helper function:

   

   The same change is needed for the `/open-folder` handler at line 114.

3. `extension-chrome/popup.js`, line 99-108 ([link](https://github.com/surge-downloader/surge/blob/8c3e76cf36bf575e2700697d4c7e798c7860c695/extension-chrome/popup.js#L99-L108)) 

   **Standalone `getHistory` path lacks auth headers**

   In extension context, `apiCall('getHistory')` routes to the background script which calls `authHeaders()` before fetching. In standalone mode (this branch), the fetch is made directly with no auth headers, unlike all other standalone calls in this switch (e.g., `cancelDownload`) which also omit auth headers — so this is consistent, but it means standalone mode will silently fail if a token is configured on the server. The same issue exists in `extension-firefox/popup.js` at the equivalent case.

4. `extension-chrome/popup.js`, line 932-937 ([link](https://github.com/surge-downloader/surge/blob/6db1bae64e35c81d63dc12ddc47b0739bf33d5c9/extension-chrome/popup.js#L932-L937)) 

   **`shouldRefreshHistory` bypasses interval for empty history**

   `historyDownloads.size === 0` conflates "never fetched" with "genuinely empty server response". When the server returns an empty history array, `historyLastFetchedAt` is updated, but on the next poll `size === 0` still evaluates to `true` — causing the 15-second `HISTORY_REFRESH_INTERVAL_MS` to be completely bypassed. Every active-downloads poll cycle in history view will re-fetch history indefinitely.

   The intent was to trigger an immediate first fetch, but the right check for that is `historyLastFetchedAt === 0`:

   

   The same bug is present in `extension-firefox/popup.js` at the equivalent function.

   **Rule Used:** What: All code changes must include tests for edge... ([source](https://app.greptile.com/review/custom-context?memory=2b22782d-3452-4d55-b059-e631b2540ce8))

5. `cmd/http_api.go`, line 122-131 ([link](https://github.com/surge-downloader/surge/blob/6db1bae64e35c81d63dc12ddc47b0739bf33d5c9/cmd/http_api.go#L122-L131)) 

   **`ErrNoDestinationPath` returns 500 instead of a 4xx**

   `ErrNoDestinationPath` falls through to the `default` branch and returns `http.StatusInternalServerError`, but this condition (the download record exists in history but its `DestPath` was never recorded) is a predictable, client-visible state — not an unexpected server fault. Returning 500 misleads callers into thinking the server crashed.

   A 422 Unprocessable Entity (request understood but the resource is in an unusable state for this operation) is semantically more accurate. Consider adding an explicit case:

   

   `TestOpenEndpoints_ReturnMappedResolveStatuses` also lacks a case for this error path, which is how it went undetected.

   **Rule Used:** What: All code changes must include tests for edge... ([source](https://app.greptile.com/review/custom-context?memory=2b22782d-3452-4d55-b059-e631b2540ce8))

6. `cmd/http_api.go`, line 98-100 ([link](https://github.com/surge-downloader/surge/blob/0bdfa451d29289b4d132b42949900dd9fa583edd/cmd/http_api.go#L98-L100)) 

   **File-not-found on disk mapped to 500**

   `utils.OpenFile` calls `os.Stat(destPath)` before dispatching to the OS; if the file was deleted after download, `os.Stat` returns a not-found error. Mapping that to `http.StatusInternalServerError` is misleading — the server is healthy, only the target file is gone.

   The same issue applies to `/open-folder` at line 119.

   Consider checking `os.IsNotExist` to return a more accurate status:

   ```go
   if err := utils.OpenFile(destPath); err != nil {
       status := http.StatusInternalServerError
       if os.IsNotExist(err) {
           status = http.StatusNotFound
       }
       http.Error(w, "Failed to open file: "+err.Error(), status)
       return
   }
   ```

   Apply the same pattern to the `/open-folder` handler.

7. `extension-chrome/popup.js`, line 211-220 ([link](https://github.com/surge-downloader/surge/blob/0bdfa451d29289b4d132b42949900dd9fa583edd/extension-chrome/popup.js#L211-L220)) 

   **Redundant client-side history sort**

   `sortDownloadsForView` re-sorts history entries by `completedAt` descending, which is exactly the order the backend already guarantees (as the `background.js` comment now correctly notes: "backend already returns history newest-first"). Since `normalizeHistoryEntry` maps `entry.completed_at` straight through to `completedAt`, this sort is a no-op on well-behaved data.

   The same redundant sort exists in `extension-firefox/popup.js` at the equivalent `sortDownloadsForView` function (line 211).

   Consider removing the history branch — the active-downloads branch still needs its status-priority sort:

   ```js
   function sortDownloadsForView(items) {
     const statusOrder = { downloading: 0, paused: 1, queued: 2, completed: 3, error: 4 };
     return items.sort((left, right) => {
       const orderLeft = statusOrder[left.status] ?? 5;
       const orderRight = statusOrder[right.status] ?? 5;
       if (orderLeft !== orderRight) return orderLeft - orderRight;
       return (right.addedAt || 0) - (left.addedAt || 0);
     });
   }
   ```

   (History view entries are already in the right order; callers in active view still benefit from the status-priority sort.)

   **Rule Used:** What: Eliminate duplicate logic, functions, or cod... ([source](https://app.greptile.com/review/custom-context?memory=f0a796a9-684f-4dfb-a5cf-8c99c16b63a1))

8. `cmd/http_api.go`, line 267-276 ([link](https://github.com/surge-downloader/surge/blob/0ca7d6dd80fb0b22ba9289d79d6bbc72b17ab1c6/cmd/http_api.go#L267-L276)) 

   **`ErrNoDestinationPath` maps to 500 Internal Server Error**

   `ErrNoDestinationPath` is not handled in `statusCodeForResolveDownloadError`, so it falls through to `http.StatusInternalServerError`. However, this error means the download entry *was found* in history but lacks a recorded destination path — that is not a server malfunction. Returning 500 misleads callers and monitoring tools into thinking something went wrong with the server itself.

   A semantically accurate alternative is `422 Unprocessable Entity` (the request was valid, the item was found, but it cannot be acted upon), or at minimum `404` to indicate the actionable resource (the file on disk) is not locatable.

9. `internal/utils/open_test.go`, line 8-45 ([link](https://github.com/surge-downloader/surge/blob/0ca7d6dd80fb0b22ba9289d79d6bbc72b17ab1c6/internal/utils/open_test.go#L8-L45)) 

   **Missing test edge case for non-existent file path**

   `TestOpenFile_Validation` covers empty-path and directory-path inputs, but `OpenFile` also calls `os.Stat(path)` for any non-empty path. A non-existent file path will return the `os.Stat` error rather than a custom validation message, and this branch is not currently exercised.

   Adding a case like the following ensures the stat-error path is covered:

   ```go
   {
       name:       "non-existent file",
       path:       filepath.Join(t.TempDir(), "does_not_exist.bin"),
       errorHints: []string{"no such file", "cannot find"},
   },
   ```

   Similarly, `TestOpenContainingFolder_Validation` could add a case for a non-existent path to verify that the `os.Stat(targetPath)` error is surfaced correctly.

   **Rule Used:** What: All code changes must include tests for edge... ([source](https://app.greptile.com/review/custom-context?memory=2b22782d-3452-4d55-b059-e631b2540ce8))

10. `cmd/http_api.go`, line 267-276 ([link](https://github.com/surge-downloader/surge/blob/e5dbe46ce8bd4d6b9cadc8d2172813e34c5ef8b6/cmd/http_api.go#L267-L276)) 

    **`ErrNoDestinationPath` unmapped — falls through to 500**

    `ErrServiceUnavailable` and `ErrDownloadNotFound` both receive their own HTTP status codes, but `ErrNoDestinationPath` is not listed in `statusCodeForResolveDownloadError`, so it silently falls through to `http.StatusInternalServerError`. That's misleading — the server isn't broken; the particular download just lacks a recorded destination path. A `422 Unprocessable Entity` (or even `404`) better communicates the semantic.

    Add an explicit case to make the intent clear and keep the three sentinel errors consistently mapped:

    **Rule Used:** # Code Review Rule: Suggest Architectural Improvem... ([source](https://app.greptile.com/review/custom-context?memory=4d82344e-33c4-4e14-bbc7-81d79af11b7e))
</details>

`````markdown
Path: extension-chrome/popup.js
Line: 459-461

Comment:
**Unreachable `> 1 week` branch in `formatETA`**

The week check at line 461 can never be reached. `seconds > 86400` (line 460) catches any value over 1 day (86 400 s), so the `seconds > 604 800` (one week) branch is dead code — durations longer than a week are always displayed as `"> 1 day"` instead.

The two guards need to be swapped so the longer threshold is evaluated first:

```suggestion
  if (seconds > 3600 * 24 * 7) return '> 1 week';
  if (seconds > 86400) return '> 1 day';
```

The same unreachable-branch bug is present in `extension-firefox/popup.js` at the equivalent lines.

---

Path: cmd/http_api.go
Line: 267-276

Comment:
**`ErrNoDestinationPath` unmapped — falls through to 500**

`ErrServiceUnavailable` and `ErrDownloadNotFound` both receive their own HTTP status codes, but `ErrNoDestinationPath` is not listed in `statusCodeForResolveDownloadError`, so it silently falls through to `http.StatusInternalServerError`. That's misleading — the server isn't broken; the particular download just lacks a recorded destination path. A `422 Unprocessable Entity` (or even `404`) better communicates the semantic.

Add an explicit case to make the intent clear and keep the three sentinel errors consistently mapped:

```suggestion
func statusCodeForResolveDownloadError(err error) int {
	switch {
	case errors.Is(err, ErrDownloadNotFound):
		return http.StatusNotFound
	case errors.Is(err, ErrServiceUnavailable):
		return http.StatusServiceUnavailable
	case errors.Is(err, ErrNoDestinationPath):
		return http.StatusUnprocessableEntity
	default:
		return http.StatusInternalServerError
	}
}
```

**Rule Used:** # Code Review Rule: Suggest Architectural Improvem... ([source](https://app.greptile.com/review/custom-context?memory=4d82344e-33c4-4e14-bbc7-81d79af11b7e))

---

Path: extension-chrome/popup.js
Line: 299-366

Comment:
**"What" comments restate the code, not the "why"**

The section labels inside `updateDownloadItem` (`// Filename`, `// Stats`, `// Status`, `// Icon`, `// Progress Bar`, `// Details`, `// Actions`) simply describe what the next few lines do, which is already obvious from the code itself. Comments should explain reasoning or non-obvious constraints, not narrate operations.

For example, the only comment here that adds value is `// Safe checks in case DOM is malformed` — it explains a defensive guard that isn't self-evident. The remaining labels should either be removed entirely or replaced with something that explains *why* (e.g., "// Compare before assigning to avoid layout thrashing on every poll tick" if that were the intent).

The same pattern exists in the equivalent `updateDownloadItem` in `extension-firefox/popup.js`.

**Rule Used:** What: Comments must explain *why* code exists, not... ([source](https://app.greptile.com/review/custom-context?memory=4e45ef13-32c2-4753-8060-a44ef767144d))

---

Path: internal/utils/open_test.go
Line: 1-82

Comment:
**No test coverage for `buildOpenCommand` platform routing**

`TestOpenFile_Validation` and `TestOpenContainingFolder_Validation` only exercise the pre-flight guards (empty path, directory path). The newly-added `buildOpenCommand` function — which selects `open`, `cmd /c start`, or `xdg-open` based on `runtime.GOOS` — has no test at all.

A table-driven test could assert the correct binary and arguments are chosen for each platform without actually executing the command:

```go
func TestBuildOpenCommand_PlatformRouting(t *testing.T) {
    cmd := buildOpenCommand("/some/path")
    // On darwin: cmd.Path ends with "open"
    // On windows: cmd.Args contains "cmd", "/c", "start"
    // On linux: cmd.Path ends with "xdg-open"
    if cmd == nil {
        t.Fatal("expected a non-nil command")
    }
}
```

Since `runtime.GOOS` is fixed at compile time in tests, you can at minimum verify the current-platform branch produces a non-nil, non-empty command with the expected argument signature.

**Rule Used:** What: All code changes must include tests for edge... ([source](https://app.greptile.com/review/custom-context?memory=2b22782d-3452-4d55-b059-e631b2540ce8))

`````

</details>

<sub>Last reviewed commit: ["trigger github workf..."](https://github.com/surge-downloader/surge/commit/e5dbe46ce8bd4d6b9cadc8d2172813e34c5ef8b6)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Rule used - # Code Review Rule: Suggest Architectural Improvem... ([source](https://app.greptile.com/review/custom-context?memory=4d82344e-33c4-4e14-bbc7-81d79af11b7e))
</details>

### DIFF
--- a/cmd/http_api_test.go
+++ b/cmd/http_api_test.go
@@ -176,7 +176,7 @@ func TestResolveDownloadDestPath(t *testing.T) {
 					"fallback": {ID: "fallback", DestPath: ""},
 				},
 				history: []types.DownloadEntry{{ID: "fallback", DestPath: "C:\\tmp\\b.bin"}},
-			},
+			}, 
 			id:       "fallback",
 			wantPath: `C:\tmp\b.bin`,
 		},

### CODE REVIEW

**Comment by FHYQ-Dong** [comment]
@SuperCoolPencil Wow, I'm pushing my commit and I see you've already done it, LOL.

**Comment by SuperCoolPencil** [comment]
Hey @FHYQ-Dong 

Thanks for the PR! History tab is a good addition to extension, if we could just cover the pathways with required tests we would be ready....

> @SuperCoolPencil Wow, I'm pushing my commit and I see you've already done it, LOL.

I'm so sorry, please feel free to undo my changes if you felt they were unnecessary, not headed in a right direction!

**Comment by FHYQ-Dong** [comment]
> Hey @FHYQ-Dong
> 
> Thanks for the PR! History tab is a good addition to extension, if we could just cover the pathways with required tests we would be ready....
> 
> > @SuperCoolPencil Wow, I'm pushing my commit and I see you've already done it, LOL.
> 
> I'm so sorry, please feel free to undo my changes if you felt they were unnecessary, not headed in a right direction!

Please let me take a few minutes to review our changes before I make the next commit...

BTW, do you have some suggestions on designing such tests? Currently I have added tests to cover the HTTP endpoint /open-file and /open-folder.

**Comment by SuperCoolPencil** [comment]
You need a new test file to cover the validation paths (empty strings, directories vs. files) without actually invoking the OS commands.

**Comment by SuperCoolPencil** [comment]
this would be a good start:

```go
package utils

import (
	"path/filepath"
	"strings"
	"testing"
)

func TestOpenFile_Validation(t *testing.T) {
	tempDir := t.TempDir()

	// 1. Empty path
	err := OpenFile("")
	if err == nil || !strings.Contains(err.Error(), "empty") {
		t.Errorf("expected error for empty path, got: %v", err)
	}

	// 2. Passing a directory to OpenFile
	err = OpenFile(tempDir)
	if err == nil || !strings.Contains(err.Error(), "is a directory") {
		t.Errorf("expected directory error, got: %v", err)
	}
}

func TestOpenContainingFolder_Validation(t *testing.T) {
	// 1. Empty path
	err := OpenContainingFolder("")
	if err == nil {
		t.Errorf("expected error for empty path")
	}

	// 2. Path resolving to "."
	err = OpenContainingFolder(".")
	if err == nil || !strings.Contains(err.Error(), "cannot resolve") {
		t.Errorf("expected cannot resolve error, got: %v", err)
	}
}
```

> (Note: You will need to make sure your actual OpenFile and OpenContainingFolder functions return these specific validation errors before calling your internal openWithSystem function).

FHYQ-Dong:
ok, let me work on it.

> this would be a good start:
> 
> ```go
> package utils
> 
> import (
> 	"path/filepath"
> 	"strings"
> 	"testing"
> )
> 
> func TestOpenFile_Validation(t *testing.T) {
> 	tempDir := t.TempDir()
> 
> 	// 1. Empty path
> 	err := OpenFile("")
> 	if err == nil || !strings.Contains(err.Error(), "empty") {
> 		t.Errorf("expected error for empty path, got: %v", err)
> 	}
> 
> 	// 2. Passing a directory to OpenFile
> 	err = OpenFile(tempDir)
> 	if err == nil || !strings.Contains(err.Error(), "is a directory") {
> 		t.Errorf("expected directory error, got: %v", err)
> 	}
> }
> 
> func TestOpenContainingFolder_Validation(t *testing.T) {
> 	// 1. Empty path
> 	err := OpenContainingFolder("")
> 	if err == nil {
> 		t.Errorf("expected error for empty path")
> 	}
> 
> 	// 2. Path resolving to "."
> 	err = OpenContainingFolder(".")
> 	if err == nil || !strings.Contains(err.Error(), "cannot resolve") {
> 		t.Errorf("expected cannot resolve error, got: %v", err)
> 	}
> }
> ```

FHYQ-Dong:
Hi @SuperCoolPencil, I've applied the suggestions provided by code review and added a test file, please review it, thx.

FHYQ-Dong:
It seems the tests fail on Windows:
```bash
=== Failed
=== FAIL: internal/core TestIntegration_PauseResume_ColdPath_StateContinuity (26.49s)
    pause_resume_integration_test.go:345: timeout waiting for status for c686a1cf-323e-49c1-b1ee-29d4474bc514; last status: status=downloading downloaded=0 total=100663296 speed=0.0000 conns=0 progress=0.000
    testing.go:1369: TempDir RemoveAll cleanup: unlinkat C:\Users\RUNNER~1\AppData\Local\Temp\TestIntegration_PauseResume_ColdPath_StateContinuity4000520582\002\cold-continuity.bin.surge: The process cannot access the file because it is being used by another process.

=== FAIL: internal/processing TestProbeServer_ReadsBodyBeforeContextCancel (2.00s)
    probe_test.go:120: ProbeServer() failed: probe request failed after retries: Get "http://127.0.0.1:52405/": proxyconnect tcp: dial tcp 127.0.0.1:52397: connectex: No connection could be made because the target machine actively refused it.
```

But this error has existed before my pr...

SuperCoolPencil:
@FHYQ-Dong 

Thanks for the PR! 🚀

FHYQ-Dong:
Report from Github Copilot which may be helpful:

## Local Reproduction and Findings

Reproduced locally (Windows) with:

```powershell
go test ./internal/processing -run "TestProbeServer_UsesConfiguredProxy|TestProbeServer_ReadsBodyBeforeContextCancel" -count=1 -v
```

Observed sequence:

1. `TestProbeServer_UsesConfiguredProxy` passes first.
2. `TestProbeServer_ReadsBodyBeforeContextCancel` then fails and attempts to use a closed local proxy port.

---

## Root Cause Analysis

### 1) The failing test calls `ProbeServer`

`TestProbeServer_ReadsBodyBeforeContextCancel` currently uses:

- `processing.ProbeServer(ctx, server.URL, "", nil)`

`ProbeServer` reads persisted settings for `Network.ProxyURL` via `resolveProxyURL -> config.LoadSettings`.

### 2) A prior test writes proxy settings

`TestProbeServer_UsesConfiguredProxy` calls `config.SaveSettings(...)` and sets `ProxyURL=proxy.URL`.

### 3) Config directory isolation is incomplete on Windows

The test only sets `XDG_CONFIG_HOME`, but on Windows `config.GetSurgeDir()` prefers `APPDATA`. As a result, proxy settings may be written to a real/shared config directory and leak into later tests.

### 4) Direct failure chain

- The proxy server is closed after the first test.
- The next test calls `ProbeServer` and reads the stale `ProxyURL`.
- The request goes through proxy and fails with `proxyconnect tcp ... actively refused`.

Conclusion: this is a **test isolation / global config contamination issue**, not a core logic bug in the body-read vs context-cancel behavior of `ProbeServer`.

---

## Fix Objective

Ensure `TestProbeServer_ReadsBodyBeforeContextCancel` only validates response-body reading and cancel timing behavior, without being affected by proxy settings or environment state.

---

## Recommended Fix (Minimal and Low Risk)

### Option A (Preferred): bypass persisted proxy config in this test

Change the failing test from `ProbeServer(...)` to `ProbeServerWithProxy(..., "")`:

- Current: `processing.ProbeServer(ctx, server.URL, "", nil)`
- Proposed: `processing.ProbeServerWithProxy(ctx, server.URL, "", nil, "")`

Benefits:

- Explicitly disables proxy usage for this test.
- Fully decouples the test from persisted config files.
- Keeps production logic unchanged and does not reduce proxy feature coverage.

### Option B (Hardening): enforce cross-platform config isolation in tests

In processing-layer tests that call `config.SaveSettings/LoadSettings`, set both:

- `t.Setenv("XDG_CONFIG_HOME", t.TempDir())`
- `t.Setenv("APPDATA", t.TempDir())`

Benefits:

- Handles Windows config path priority correctly.
- Prevents cross-test contamination and local machine config interference.
- Improves CI stability.

---

## Not Recommended

### Modifying production proxy/retry logic

Not recommended. The issue is test isolation, not production behavior; changing runtime logic introduces unnecessary risk.

---

## Implementation Steps

1. Update `internal/processing/probe_test.go`:
   - Change `TestProbeServer_ReadsBodyBeforeContextCancel` to call `ProbeServerWithProxy(..., "")`.
2. (Recommended) Add `APPDATA` isolation setup in `TestProbeServer_UsesConfiguredProxy`.
3. Run targeted regression:
   - `go test ./internal/processing -run "TestProbeServer_UsesConfiguredProxy|TestProbeServer_ReadsBodyBeforeContextCancel" -count=1 -v`
4. Run probe-related test subset:
   - `go test ./internal/processing -run Probe -count=1`
5. If green, run the equivalent CI test subset.

---

## Acceptance Criteria

- `TestProbeServer_ReadsBodyBeforeContextCancel` passes consistently (local + CI).
- No more `proxyconnect tcp ... actively refused` failures.
- `TestProbeServer_UsesConfiguredProxy` still verifies proxy routing (`proxyHits > 0`, `directHits == 0`).
- No production behavior changes in `internal/processing/probe.go`.

---

## Risk and Rollback

- Risk level: Low (test-only changes).
- Rollback: revert `internal/processing/probe_test.go`.

SuperCoolPencil:
Thanks so much :)

PR #293, opened 2026-03-23
TUI: replace manual tab state with paginator-backed models

### LINKED ISSUE #278: TUI: Replace custom tab bar logic with Bubbles paginator-backed state
## Summary
Replace custom tab-bar state/rendering with Bubbles `paginator.Model`-driven tab state.

## Current State
`internal/tui/components/tab_bar.go` manually builds and styles tab labels/counters.

## Why
- Reduce custom tab rendering/state logic.
- Reuse maintained Bubbles behavior for pagination-like tab selection.

## Scope
- Dashboard tabs (Queued/Active/Done).
- Settings/category tab-like selectors where appropriate.
- Preserve current hotkeys and visible labels/counters.

## Acceptance Criteria
- Tab switching behavior unchanged.
- Visual styling remains consistent with current theme.
- Existing tab-related tests pass or are updated.

standard-librarian:
## Summary
- replace manual dashboard tab index handling with bubbles paginator Model-backed state
- replace settings category tab index handling with bubbles paginator Model-backed state
- preserve existing keybindings, labels/counters, and tab styling
- keep compatibility mirrors for existing paths/tests while routing behavior through paginator helpers

## Validation
- go test ./...
- go build ./...

Closes #278

<h3>Greptile Summary</h3>

This PR replaces direct `activeTab` / `SettingsActiveTab` field mutations throughout the TUI with `paginator.Model`-backed helpers (`currentDashboardTab`, `setDashboardTab`, `nextDashboardTab`, `currentSettingsTab`, etc.), aiming to centralise tab-state management and lay groundwork for bubbletea-native paginator integration. Companion tests cover the new helpers, clamping, wrap-around, diverged-source rendering, and the category-count-shrink reinit path.

- **`ensureDashboardTabs` / `ensureSettingsTabs`** now early-return when the paginator is already correctly initialised, eliminating the per-frame reallocation flagged in a previous round.
- **Dual-source pattern** (`paginator.Page` + a legacy mirror field) is preserved intentionally for backward compatibility; every mutation helper keeps the mirror in sync.
- **`view_settings.go`** now reads `activeSettingsTab` from `currentSettingsTab()` and the right-panel unit uses the stateless `getSettingUnitForKey(meta.Key)`, cleanly removing the prior dual-source right-panel inconsistency.
- **DRY violation in `updateSettings` Edit block**: `activeSettingsTab` is computed at line 103 via `m.currentSettingsTab(categoryCount)` but `m.currentSettingsTab(categoryCount)` is called again on line 126 for `currentCategory` — the existing variable should be reused.
- **`categoryCount` parameter threading**: every settings-tab helper requires callers to supply `len(config.CategoryOrder())` manually; since `config.CategoryOrder()` is deterministic and already called at every call site, internalising this inside the helpers would simplify the API and eliminate the risk of a caller passing a stale count.
- **`TestSettingsHelpersUsePaginatorWhenSourcesDiverge`** hardcodes that row 0 of General is `"default_download_dir"` with type `"string"` — the same position-sensitive fragility seen elsewhere in the test suite; the row should be located by key at runtime.

- Safe to merge for existing functionality, but open design questions from prior rounds (paginator.Update() never called, dual-source mirrors) remain unresolved and the API has a few rough edges flagged in this round.
- All existing tab-switching behaviour is correctly preserved through the paginator helpers, and the early-return guards eliminate the per-frame reallocation. However, the paginator component is still not wired into the bubbletea message loop (Update() never called), meaning it functions purely as a plain struct — adding the dependency cost without delivering the paginator's intended event-driven value. The categoryCount threading creates a caller-side error surface, and a DRY violation was introduced in the Edit block. The test suite is meaningfully expanded but contains position-sensitive hardcoded assertions.
- Pay close attention to `internal/tui/model.go` (paginator helper API design and the uncalled Update() concern) and `internal/tui/paginator_tabs_test.go` (position-sensitive row assertions).

| Filename | Overview |
|----------|----------|
| internal/tui/model.go | Core of the PR — adds paginator.Model fields and all tab-management helpers; early-return guards in ensureDashboardTabs/ensureSettingsTabs are now correct, but the categoryCount parameter on every settings helper creates a fragile caller API, and paginator.Update() remains uncalled. |
| internal/tui/update_settings.go | All direct SettingsActiveTab mutations replaced with paginator helpers; one DRY violation remains — activeSettingsTab computed at line 103 but currentSettingsTab called again at line 126 for currentCategory. |
| internal/tui/paginator_tabs_test.go | Good breadth of new tests including diverged-source scenarios; TestSettingsHelpersUsePaginatorWhenSourcesDiverge hardcodes row 0 = "default_download_dir" which is position-sensitive. |
| internal/tui/view_settings.go | viewSettings now reads from paginator via currentSettingsTab; getSettingUnit replaced by stateless getSettingUnitForKey keyed on meta.Key from the paginator-derived category — dual-source right-panel bug cleanly resolved. |
| internal/tui/update_dashboard.go | All activeTab mutations correctly routed through setDashboardTab/nextDashboardTab; no issues found. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Key Event / Update call] --> B{Which state?}
    B -->|DashboardState| C[updateDashboard]
    B -->|SettingsState| D[updateSettings]

    C --> C1[setDashboardTab / nextDashboardTab]
    C1 --> C2[ensureDashboardTabs]
    C2 --> C3{dashboardTabs\nalready init?}
    C3 -->|Yes - early return| C4[dashboardTabs.Page]
    C3 -->|No - reinit| C5[newTabPaginator\ndashboardTabCount=3]
    C5 --> C6[seed from activeTab mirror]
    C6 --> C4
    C4 --> C7[activeTab mirror synced]
    C7 --> C8[View: renderTabs currentDashboardTab]

    D --> D1[setSettingsTab / next/prevSettingsTab]
    D1 --> D2[ensureSettingsTabs categoryCount]
    D2 --> D3{settingsTabs\nalready init with\nsame totalPages?}
    D3 -->|Yes - early return| D4[settingsTabs.Page]
    D3 -->|No - reinit| D5[newTabPaginator\ncategoryCount]
    D5 --> D6[seed from SettingsActiveTab mirror]
    D6 --> D4
    D4 --> D7[SettingsActiveTab mirror synced]
    D7 --> D8[viewSettings: currentSettingsTab]
    D8 --> D9[getCurrentSettingMeta\ngetSettingsCount\ngetCurrentSettingType\nall route through currentSettingsTab]
```

<details open><summary><h3>Comments Outside Diff (6)</h3></summary>

1. `internal/tui/model.go`, line 209-218 ([link](https://github.com/surge-downloader/surge/blob/82e7951d9aa5107de8116ee697618b63b2fcc888/internal/tui/model.go#L209-L218)) 

   <a href="#">P1</a> **Paginator is rebuilt unconditionally on every read**

   `ensureDashboardTabs()` calls `newTabPaginator(dashboardTabCount)` and overwrites `m.dashboardTabs` on *every invocation*, including every call to `currentDashboardTab()`, which in turn is called from `View()`, `updateListTitle()`, and `getFilteredDownloads()` — meaning the paginator is fully recreated on every render cycle.

   After `InitialRootModel` runs, `m.dashboardTabs.PerPage` and `m.dashboardTabs.TotalPages` will always satisfy the guard condition (`>= 1`), so the `m.activeTab` fallback is dead in practice. The result is:
   1. Every read of the current tab allocates a new `paginator.Model`
   2. Any state that the paginator accumulates (e.g. from future `Update` calls) is silently discarded

   The simplest fix is to only initialise the paginator once (it already is, in `InitialRootModel`) and remove `ensureDashboardTabs` entirely, reading `m.dashboardTabs.Page` directly:

   ```go
   func (m *RootModel) currentDashboardTab() int {
       return clampTabIndex(m.dashboardTabs.Page, dashboardTabCount)
   }
   ```

   The same problem exists in `ensureSettingsTabs` / `currentSettingsTab`.

   `````markdown
   Path: internal/tui/model.go
   Line: 209-218

   Comment:
   **Paginator is rebuilt unconditionally on every read**

   `ensureDashboardTabs()` calls `newTabPaginator(dashboardTabCount)` and overwrites `m.dashboardTabs` on *every invocation*, including every call to `currentDashboardTab()`, which in turn is called from `View()`, `updateListTitle()`, and `getFilteredDownloads()` — meaning the paginator is fully recreated on every render cycle.

   After `InitialRootModel` runs, `m.dashboardTabs.PerPage` and `m.dashboardTabs.TotalPages` will always satisfy the guard condition (`>= 1`), so the `m.activeTab` fallback is dead in practice. The result is:
   1. Every read of the current tab allocates a new `paginator.Model`
   2. Any state that the paginator accumulates (e.g. from future `Update` calls) is silently discarded

   The simplest fix is to only initialise the paginator once (it already is, in `InitialRootModel`) and remove `ensureDashboardTabs` entirely, reading `m.dashboardTabs.Page` directly:

   ```go
   func (m *RootModel) currentDashboardTab() int {
       return clampTabIndex(m.dashboardTabs.Page, dashboardTabCount)
   }
   ```

   The same problem exists in `ensureSettingsTabs` / `currentSettingsTab`.

   `````
   </details>

2. `internal/tui/model.go`, line 231-239 ([link](https://github.com/surge-downloader/surge/blob/82e7951d9aa5107de8116ee697618b63b2fcc888/internal/tui/model.go#L231-L239)) 

   <a href="#">P2</a> **Manual wrap-around re-implements paginator's built-in navigation**

   `nextDashboardTab()`, `nextSettingsTab()`, and `prevSettingsTab()` all hand-roll the wrap-around logic:

   ```go
   if m.dashboardTabs.Page >= m.dashboardTabs.TotalPages-1 {
       m.dashboardTabs.Page = 0
   } else {
       m.dashboardTabs.Page++
   }
   ```

   The `paginator.Model` from `charm.land/bubbles/v2/paginator` exposes `NextPage()` and `PrevPage()` methods that already handle boundary and wrap-around behaviour. Delegating to those methods would eliminate this custom logic and let the library own that responsibility — which is exactly why it was introduced.

   The same pattern applies to `nextSettingsTab` (line 267) and `prevSettingsTab` (line 277).

   **Rule Used:** What: Prefer built-in or existing library function... ([source](https://app.greptile.com/review/custom-context?memory=545c1730-b3a2-4153-a981-7f3f23c2e226))
   
   `````markdown
   Path: internal/tui/model.go
   Line: 231-239
   
   Comment:
   **Manual wrap-around re-implements paginator's built-in navigation**
   
   `nextDashboardTab()`, `nextSettingsTab()`, and `prevSettingsTab()` all hand-roll the wrap-around logic:
   
   ```go
   if m.dashboardTabs.Page >= m.dashboardTabs.TotalPages-1 {
       m.dashboardTabs.Page = 0
   } else {
       m.dashboardTabs.Page++
   }
   ```
   
   The `paginator.Model` from `charm.land/bubbles/v2/paginator` exposes `NextPage()` and `PrevPage()` methods that already handle boundary and wrap-around behaviour. Delegating to those methods would eliminate this custom logic and let the library own that responsibility — which is exactly why it was introduced.
   
   The same pattern applies to `nextSettingsTab` (line 267) and `prevSettingsTab` (line 277).
   
   **Rule Used:** What: Prefer built-in or existing library function... ([source](https://app.greptile.com/review/custom-context?memory=545c1730-b3a2-4153-a981-7f3f23c2e226))
   
   `````
   </details>

3. `internal/tui/update.go`, line 494-497 ([link](https://github.com/surge-downloader/surge/blob/82e7951d9aa5107de8116ee697618b63b2fcc888/internal/tui/update.go#L494-L497)) 

   <a href="#">P1</a> **`ctrl+c` in `QuitConfirmState` cancels instead of confirming — double-press no longer quits**

   Before this PR, `ctrl+c` immediately initiated shutdown. Now it opens the quit confirmation modal. The `Cancel` binding (`esc`, `ctrl+c`, `ctrl+q`) maps `ctrl+c` to *dismissing* the modal and returning to dashboard.

   A user who reflexively hits `ctrl+c` twice will be left confused: the first press opens the dialog, the second press closes it with no quit — an infinite cycle. This is a meaningful UX regression from the previous behaviour.

   Consider either:
   - Removing `ctrl+c` from `Cancel` (keep it only on `esc`) so that pressing it inside the modal falls through as an unhandled key, or
   - Mapping `ctrl+c` to `confirmQuit()` inside `QuitConfirmState` so the second `ctrl+c` completes the quit.

   There is also a test `TestQuitConfirm_CtrlCCancels` that asserts this dismissal behaviour — that test would need updating if the binding changes.

   `````markdown
   Path: internal/tui/update.go
   Line: 494-497

   Comment:
   **`ctrl+c` in `QuitConfirmState` cancels instead of confirming — double-press no longer quits**

   Before this PR, `ctrl+c` immediately initiated shutdown. Now it opens the quit confirmation modal. The `Cancel` binding (`esc`, `ctrl+c`, `ctrl+q`) maps `ctrl+c` to *dismissing* the modal and returning to dashboard.

   A user who reflexively hits `ctrl+c` twice will be left confused: the first press opens the dialog, the second press closes it with no quit — an infinite cycle. This is a meaningful UX regression from the previous behaviour.

   Consider either:
   - Removing `ctrl+c` from `Cancel` (keep it only on `esc`) so that pressing it inside the modal falls through as an unhandled key, or
   - Mapping `ctrl+c` to `confirmQuit()` inside `QuitConfirmState` so the second `ctrl+c` completes the quit.

   There is also a test `TestQuitConfirm_CtrlCCancels` that asserts this dismissal behaviour — that test would need updating if the binding changes.

   `````
   </details>

4. `internal/tui/view_settings.go`, line 487-498 ([link](https://github.com/surge-downloader/surge/blob/72fb430d2cd22aca5a57f320273436722c820854/internal/tui/view_settings.go#L487-L498)) 

   <a href="#">P2</a> **View-layer helpers read `SettingsActiveTab` mirror inconsistently**

   `getCurrentSettingMeta()`, `getSettingsCount()` (line 514), and `getCurrentSettingKey()` (line 463) all read `m.SettingsActiveTab` directly. Meanwhile `updateSettings()` reads the active tab exclusively through `m.currentSettingsTab(categoryCount)`. This means the update path and the query-helper path draw from different fields, even though both are meant to reflect the same value.

   Since every mutation flows through the helper methods that keep both fields in sync, there is no current bug. However, the split makes the code harder to reason about and creates a maintenance trap: a future contributor modifying only the paginator path (without updating the mirror) will see the update logic behave correctly while the helpers silently return stale data.

   Consider routing all reads through a single field — either use `m.currentSettingsTab()` consistently everywhere, or make the paginator-backed helpers the mirror (and remove `SettingsActiveTab` from direct use in view code).

   `````markdown
   Path: internal/tui/view_settings.go
   Line: 487-498

   Comment:
   **View-layer helpers read `SettingsActiveTab` mirror inconsistently**

   `getCurrentSettingMeta()`, `getSettingsCount()` (line 514), and `getCurrentSettingKey()` (line 463) all read `m.SettingsActiveTab` directly. Meanwhile `updateSettings()` reads the active tab exclusively through `m.currentSettingsTab(categoryCount)`. This means the update path and the query-helper path draw from different fields, even though both are meant to reflect the same value.

   Since every mutation flows through the helper methods that keep both fields in sync, there is no current bug. However, the split makes the code harder to reason about and creates a maintenance trap: a future contributor modifying only the paginator path (without updating the mirror) will see the update logic behave correctly while the helpers silently return stale data.

   Consider routing all reads through a single field — either use `m.currentSettingsTab()` consistently everywhere, or make the paginator-backed helpers the mirror (and remove `SettingsActiveTab` from direct use in view code).

   `````
   </details>

5. `internal/tui/view_settings.go`, line 481-516 ([link](https://github.com/surge-downloader/surge/blob/d73f2f39745b500db0a8e4a437da23c87f79c261/internal/tui/view_settings.go#L481-L516)) 

   <a href="#">P1</a> **`getCurrentSettingMeta` and `getSettingsCount` bypass the paginator**

   Both helpers still read `m.SettingsActiveTab` directly rather than going through `currentSettingsTab()`. This is the one remaining dual-source inconsistency in the PR: `viewSettings` was updated to call `currentSettingsTab(categoryCount)`, but these two callers were not.

   The concrete failure path is in the `Down` key handler in `updateSettings.go` (line 93):

   ```go
   maxRow := m.getSettingsCount() - 1
   ```

   This runs _before_ any `currentSettingsTab()` call on that code path. If `m.SettingsActiveTab` ever holds an out-of-range value while `m.settingsTabs.Page` is clamped correctly (e.g., a model constructed externally or by a test with `SettingsActiveTab = 99`), `getSettingsCount()` hits the `m.SettingsActiveTab >= categoryCount` branch and returns `0`. `maxRow` becomes `-1`, so `Down` navigation silently does nothing even though the paginator would have served a valid tab.

   `getCurrentSettingMeta()` has the same pattern: an OOB `SettingsActiveTab` causes it to return `nil` early, making `getCurrentSettingType()` always return `"string"` and `getCurrentSettingKey()` return `""` — both wrong.

   ```go
   // getCurrentSettingMeta
   activeSettingsTab := m.SettingsActiveTab  // should call currentSettingsTab(categoryCount)

   // getSettingsCount
   if categoryCount > 0 && m.SettingsActiveTab >= 0 && m.SettingsActiveTab < categoryCount {  // same issue
   ```

   These two functions should resolve their active tab via `currentSettingsTab(categoryCount)` for the same reason `viewSettings` was updated to do so.

   `````markdown
   Path: internal/tui/view_settings.go
   Line: 481-516

   Comment:
   **`getCurrentSettingMeta` and `getSettingsCount` bypass the paginator**

   Both helpers still read `m.SettingsActiveTab` directly rather than going through `currentSettingsTab()`. This is the one remaining dual-source inconsistency in the PR: `viewSettings` was updated to call `currentSettingsTab(categoryCount)`, but these two callers were not.

   The concrete failure path is in the `Down` key handler in `updateSettings.go` (line 93):

   ```go
   maxRow := m.getSettingsCount() - 1
   ```

   This runs _before_ any `currentSettingsTab()` call on that code path. If `m.SettingsActiveTab` ever holds an out-of-range value while `m.settingsTabs.Page` is clamped correctly (e.g., a model constructed externally or by a test with `SettingsActiveTab = 99`), `getSettingsCount()` hits the `m.SettingsActiveTab >= categoryCount` branch and returns `0`. `maxRow` becomes `-1`, so `Down` navigation silently does nothing even though the paginator would have served a valid tab.

   `getCurrentSettingMeta()` has the same pattern: an OOB `SettingsActiveTab` causes it to return `nil` early, making `getCurrentSettingType()` always return `"string"` and `getCurrentSettingKey()` return `""` — both wrong.

   ```go
   // getCurrentSettingMeta
   activeSettingsTab := m.SettingsActiveTab  // should call currentSettingsTab(categoryCount)

   // getSettingsCount
   if categoryCount > 0 && m.SettingsActiveTab >= 0 && m.SettingsActiveTab < categoryCount {  // same issue
   ```

   These two functions should resolve their active tab via `currentSettingsTab(categoryCount)` for the same reason `viewSettings` was updated to do so.

   `````
   </details>

6. `internal/tui/view_settings.go`, line 132 ([link](https://github.com/surge-downloader/surge/blob/1de8895f18f237cc36cb9c2c59c16a0e5b668379/internal/tui/view_settings.go#L132)) 

   <a href="#">P2</a> **Right-panel metadata uses mirror, not paginator**

   `viewSettings()` computes `activeSettingsTab` from the paginator (line 54) and uses it to select `settingsMeta` for the left panel. However, `m.getSettingUnit()` on line 132 calls `getCurrentSettingMeta()`, which reads `m.SettingsActiveTab` (the mirror) directly — not `activeSettingsTab`. In a diverged state (as manually constructed in `TestViewSettingsUsesPaginatorWhenSourcesDiverge`), the left panel would render General's settings list while the right panel's unit suffix and description would be derived from a different category (e.g., Performance), producing a mismatched display.

   The `TestViewSettingsUsesPaginatorWhenSourcesDiverge` test does not catch this because it only asserts on the presence/absence of setting *names* (left-column content), not on the unit or description in the right column. A simple fix is to pass `activeSettingsTab` into `getCurrentSettingMeta` (or introduce a parameter-aware variant) so the right panel reads from the same source as the left panel:

   ```go
   // current (uses mirror via getCurrentSettingMeta → m.SettingsActiveTab)
   unit := m.getSettingUnit()

   // safer: derive unit from the paginator-resolved tab directly
   unit := getSettingUnitForCategory(categories[activeSettingsTab], m.SettingsSelectedRow)
   ```

   `````markdown
   Path: internal/tui/view_settings.go
   Line: 132

   Comment:
   **Right-panel metadata uses mirror, not paginator**

   `viewSettings()` computes `activeSettingsTab` from the paginator (line 54) and uses it to select `settingsMeta` for the left panel. However, `m.getSettingUnit()` on line 132 calls `getCurrentSettingMeta()`, which reads `m.SettingsActiveTab` (the mirror) directly — not `activeSettingsTab`. In a diverged state (as manually constructed in `TestViewSettingsUsesPaginatorWhenSourcesDiverge`), the left panel would render General's settings list while the right panel's unit suffix and description would be derived from a different category (e.g., Performance), producing a mismatched display.

   The `TestViewSettingsUsesPaginatorWhenSourcesDiverge` test does not catch this because it only asserts on the presence/absence of setting *names* (left-column content), not on the unit or description in the right column. A simple fix is to pass `activeSettingsTab` into `getCurrentSettingMeta` (or introduce a parameter-aware variant) so the right panel reads from the same source as the left panel:

   ```go
   // current (uses mirror via getCurrentSettingMeta → m.SettingsActiveTab)
   unit := m.getSettingUnit()

   // safer: derive unit from the paginator-resolved tab directly
   unit := getSettingUnitForCategory(categories[activeSettingsTab], m.SettingsSelectedRow)
   ```

   `````
   </details>
</details>

`````markdown
Path: internal/tui/update_settings.go
Line: 126

Comment:
**Redundant `currentSettingsTab` call discards already-computed result**

`activeSettingsTab` was computed two lines earlier at line 103 and is still in scope, but `m.currentSettingsTab(categoryCount)` is called again here to compute `currentCategory`. The first call's result should be reused — this is the kind of duplication rule `f0a796a9` targets.

```suggestion
		currentCategory := categories[activeSettingsTab]
```

**Rule Used:** What: Eliminate duplicate logic, functions, or cod... ([source](https://app.greptile.com/review/custom-context?memory=f0a796a9-684f-4dfb-a5cf-8c99c16b63a1))

---

Path: internal/tui/model.go
Line: 549-578

Comment:
**Architectural: `categoryCount` parameter threading creates an error-prone API**

Every settings-tab helper — `currentSettingsTab`, `setSettingsTab`, `nextSettingsTab`, `prevSettingsTab`, `ensureSettingsTabs` — requires callers to supply `categoryCount` manually. In practice every call site computes `len(config.CategoryOrder())` immediately before calling, which means this data is already globally accessible:

```go
// All call sites look like this:
m.currentSettingsTab(len(config.CategoryOrder()))
m.nextSettingsTab(len(config.CategoryOrder()))
```

Since `config.CategoryOrder()` is already called multiple times per render frame in `viewSettings` and `updateSettings`, the per-call cost is not a concern. Having the helpers fetch the count internally would eliminate the parameter from every call site, making it impossible for callers to accidentally pass a stale count that triggers unexpected paginator reinitialisation:

```go
func (m *RootModel) currentSettingsTab() int {
    return m.currentSettingsTabN(len(config.CategoryOrder()))
}
// keep the N-variant for the one place that already has the slice
```

This would also remove the need for callers in `updateSettings` to declare `categoryCount` at the top of the function just to thread it through.

**Rule Used:** # Code Review Rule: Suggest Architectural Improvem... ([source](https://app.greptile.com/review/custom-context?memory=4d82344e-33c4-4e14-bbc7-81d79af11b7e))

---

Path: internal/tui/paginator_tabs_test.go
Line: 233-241

Comment:
**Hardcoded key and type for row 0 create fragile position-dependent assertions**

The test asserts `getCurrentSettingKey() == "default_download_dir"` and `getCurrentSettingType() == "string"` for `SettingsSelectedRow = 0`, assuming `default_download_dir` is always the first entry in the General category. If a setting is ever inserted before it in `GetSettingsMetadata()["General"]`, this test will silently assert the wrong property — or produce a misleading failure.

The same approach used elsewhere in the codebase to look up a row by key would make this robust:

```go
generalMeta := config.GetSettingsMetadata()["General"]
row := -1
for i, entry := range generalMeta {
    if entry.Key == "default_download_dir" {
        row = i
        break
    }
}
if row == -1 {
    t.Fatal("default_download_dir not found in General settings")
}
m.SettingsSelectedRow = row
```

Then assert `getCurrentSettingKey() == "default_download_dir"` and `getCurrentSettingType() == generalMeta[row].Type` instead of the hardcoded `"string"`.

`````

</details>

<sub>Reviews (11): Last reviewed commit: ["test(tui): avoid terminal queries in pag..."](https://github.com/surge-downloader/surge/commit/84f768443b99579fbe20f53f35acca487d3e84ad) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=25997744)</sub>

> Greptile also left **2 inline comments** on this PR.

<details><summary><h4>Context used (3)</h4></summary>

- Rule used - What: Eliminate duplicate logic, functions, or cod... ([source](https://app.greptile.com/review/custom-context?memory=f0a796a9-684f-4dfb-a5cf-8c99c16b63a1))
- Rule used - # Code Review Rule: Suggest Architectural Improvem... ([source](https://app.greptile.com/review/custom-context?memory=4d82344e-33c4-4e14-bbc7-81d79af11b7e))
</details>

### DIFF
--- a/internal/tui/paginator_tabs_test.go
+++ b/internal/tui/paginator_tabs_test.go
@@ -195,10 +195,12 @@ func TestCurrentSettingsTabClampsOOBActiveTab(t *testing.T) {
 }
 
 func TestViewSettingsUsesPaginatorWhenSourcesDiverge(t *testing.T) {
-	m := InitialRootModel(1701, "test-version", nil, nil, false)
-	m.Settings = config.DefaultSettings()
-	m.width = 120
-	m.height = 34
+	m := RootModel{
+		Settings: config.DefaultSettings(),
+		keys:     Keys,
+		width:    120,
+		height:   34,
+	}
 	m.settingsTabs = newTabPaginator(len(config.CategoryOrder()))
 	m.settingsTabs.Page = 0  // General
 	m.SettingsActiveTab = 2 // Performance
@@ -219,8 +221,10 @@ func TestViewSettingsUsesPaginatorWhenSourcesDiverge(t *testing.T) {
 }
 
 func TestSettingsHelpersUsePaginatorWhenSourcesDiverge(t *testing.T) {
-	m := InitialRootModel(1701, "test-version", nil, nil, false)
-	m.Settings = config.DefaultSettings()
+	m := RootModel{
+		Settings: config.DefaultSettings(),
+		keys:     Keys,
+	}
 	m.settingsTabs = newTabPaginator(len(config.CategoryOrder()))
 	m.settingsTabs.Page = 0  // General
 	m.SettingsActiveTab = 2 // Performance


standard-librarian:
@SuperCoolPencil addressed the Greptile review comments. Ready for your review whenever you get a chance!​​​​​​​​​​​​​​​​

PR #311, opened 2026-04-01
fix(config): fall back to defaults when settings.json is corrupt

### LINKED ISSUE #310: bug: LoadSettings returns error on corrupt JSON — prevents application from starting
## Description

`LoadSettings()` in `internal/config/settings.go` returns an error when the settings file contains invalid JSON:

```go
settings := DefaultSettings()
if err := json.Unmarshal(data, settings); err != nil {
    return nil, err  // ← app cannot start
}
```

### Problem

If `settings.json` gets corrupted (e.g., crash during `SaveSettings()`, disk full, manual edit with syntax error), the application fails to start entirely. The user has no way to recover except manually deleting or fixing the file.

### Expected behavior

The application should fall back to `DefaultSettings()` and log a warning when the settings file is corrupt. This matches the existing behavior for a missing file (which already returns defaults).

### Fix

Replace the error return with a log warning + `DefaultSettings()` fallback:

```go
if err := json.Unmarshal(data, settings); err != nil {
    log.Printf("Warning: corrupt settings file %s: %v — using defaults", path, err)
    return DefaultSettings(), nil
}
```

### Impact

Users who experience settings corruption are currently locked out of the application. This fix ensures Surge can always start.

LuisMIguelFurlanettoSousa:
## Summary

Fixes #310

- `LoadSettings()` now falls back to `DefaultSettings()` with a log warning when `settings.json` contains invalid JSON, instead of returning an error that prevents the application from starting

## Changes

`internal/config/settings.go`:
- On `json.Unmarshal` failure: log warning with file path and error, return `DefaultSettings()` instead of `nil, err`
- Added `log` import

## Test plan

- [x] `go build ./...` compiles successfully
- [x] `go test ./internal/config/...` passes
- [ ] Manual: corrupt settings.json with invalid JSON, verify Surge starts with defaults and logs warning

<h3>Greptile Summary</h3>

This PR changes `LoadSettings()` to fall back to `DefaultSettings()` instead of returning an error when `settings.json` contains invalid JSON, preventing the application from failing to start due to a corrupt config file. Three integration tests covering corrupt, truncated, and empty file scenarios are also added.

**Key changes:**
- `json.Unmarshal` failure now returns `DefaultSettings(), nil` instead of `nil, err`, making the app resilient to config corruption
- A `utils.Debug(...)` call is used to log the warning — however, `utils.Debug` is a no-op unless verbose mode is explicitly enabled, meaning end-users will never see the warning in normal operation
- Three new integration tests validate the fallback behavior using `XDG_CONFIG_HOME` to redirect the settings path
- New test functions use Portuguese-language section comments (`// Cenário`, `// Ação`, `// Validação`) inconsistent with the English-only conventions in the rest of the file

- Safe to merge after fixing the warning mechanism — the fallback logic itself is correct but the warning is silently suppressed for all normal users.
- The fallback-to-defaults behavior is sound and the new tests are solid. The P1 issue is that the corruption warning uses utils.Debug(), which requires verbose mode to be enabled — in practice the warning is invisible to end-users, defeating the stated purpose of notifying them about the corrupt file. Switching to the standard log package or stderr output is a one-line fix.
- internal/config/settings.go line 176 — the utils.Debug call should be replaced with a mechanism that emits the warning unconditionally.

| Filename | Overview |
|----------|----------|
| internal/config/settings.go | Adds fallback-to-defaults on corrupt JSON, but uses utils.Debug() which is a no-op when verbose mode is off — the intended user-visible warning is silently dropped in normal usage. |
| internal/config/settings_test.go | Adds three well-structured integration tests for corrupt/truncated/empty settings files; minor issue with non-English section comments inconsistent with the rest of the test file. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[LoadSettings called] --> B[os.ReadFile settings.json]
    B -->|os.IsNotExist| C[return DefaultSettings, nil]
    B -->|other I/O error| D[return nil, err]
    B -->|success| E[json.Unmarshal into DefaultSettings]
    E -->|unmarshal error| F[utils.Debug warning\n⚠️ only emits if verbose=true]
    F --> G[return DefaultSettings, nil]
    E -->|success| H[return merged Settings, nil]
```

<details open><summary><h3>Comments Outside Diff (5)</h3></summary>

1. `internal/config/settings_test.go`, line 226-247 ([link](https://github.com/surge-downloader/surge/blob/7d4bdd86750bc42592be6e503ae0a0a561bd47db/internal/config/settings_test.go#L226-L247)) 

   <a href="#">P2</a> **`TestLoadSettings_CorruptedJSON` doesn't test `LoadSettings()` at all**

   The test manually calls `json.Unmarshal` on a file in a temp directory — it never invokes `LoadSettings()`. This means the new fallback behaviour (logging a warning and returning `DefaultSettings()` instead of an error) is not exercised by any test. Someone could revert line 175–176 to `return nil, err` and every existing test would still pass.

   The other `TestLoadSettings_Real*` tests show the right pattern: set `XDG_CONFIG_HOME` to a temp dir so `GetSettingsPath()` resolves inside it, write a file, then call `LoadSettings()`. A test covering this case would look something like:

   ```go
   func TestLoadSettings_CorruptedJSON_Fallback(t *testing.T) {
       t.Setenv("XDG_CONFIG_HOME", t.TempDir())

       // Write corrupt JSON to the resolved settings path
       if err := os.MkdirAll(filepath.Dir(GetSettingsPath()), 0o755); err != nil {
           t.Fatalf("mkdir: %v", err)
       }
       if err := os.WriteFile(GetSettingsPath(), []byte("{invalid json"), 0o644); err != nil {
           t.Fatalf("write: %v", err)
       }

       settings, err := LoadSettings()
       if err != nil {
           t.Fatalf("LoadSettings should not return an error for corrupt JSON, got: %v", err)
       }
       if settings == nil {
           t.Fatal("LoadSettings returned nil settings for corrupt JSON")
       }
       // Verify we got sensible defaults, not a zero-value struct
       if settings.Network.MaxConnectionsPerHost <= 0 {
           t.Errorf("Expected default MaxConnectionsPerHost, got %d", settings.Network.MaxConnectionsPerHost)
       }
   }
   ```

   **Rule Used:** What: All code changes must include tests for edge... ([source](https://app.greptile.com/review/custom-context?memory=2b22782d-3452-4d55-b059-e631b2540ce8))
   
   `````markdown
   Path: internal/config/settings_test.go
   Line: 226-247
   
   Comment:
   **`TestLoadSettings_CorruptedJSON` doesn't test `LoadSettings()` at all**
   
   The test manually calls `json.Unmarshal` on a file in a temp directory — it never invokes `LoadSettings()`. This means the new fallback behaviour (logging a warning and returning `DefaultSettings()` instead of an error) is not exercised by any test. Someone could revert line 175–176 to `return nil, err` and every existing test would still pass.
   
   The other `TestLoadSettings_Real*` tests show the right pattern: set `XDG_CONFIG_HOME` to a temp dir so `GetSettingsPath()` resolves inside it, write a file, then call `LoadSettings()`. A test covering this case would look something like:
   
   ```go
   func TestLoadSettings_CorruptedJSON_Fallback(t *testing.T) {
       t.Setenv("XDG_CONFIG_HOME", t.TempDir())
   
       // Write corrupt JSON to the resolved settings path
       if err := os.MkdirAll(filepath.Dir(GetSettingsPath()), 0o755); err != nil {
           t.Fatalf("mkdir: %v", err)
       }
       if err := os.WriteFile(GetSettingsPath(), []byte("{invalid json"), 0o644); err != nil {
           t.Fatalf("write: %v", err)
       }
   
       settings, err := LoadSettings()
       if err != nil {
           t.Fatalf("LoadSettings should not return an error for corrupt JSON, got: %v", err)
       }
       if settings == nil {
           t.Fatal("LoadSettings returned nil settings for corrupt JSON")
       }
       // Verify we got sensible defaults, not a zero-value struct
       if settings.Network.MaxConnectionsPerHost <= 0 {
           t.Errorf("Expected default MaxConnectionsPerHost, got %d", settings.Network.MaxConnectionsPerHost)
       }
   }
   ```
   
   **Rule Used:** What: All code changes must include tests for edge... ([source](https://app.greptile.com/review/custom-context?memory=2b22782d-3452-4d55-b059-e631b2540ce8))
   
   `````
   </details>

2. `internal/config/settings.go`, line 176 ([link](https://github.com/surge-downloader/surge/blob/9c8e594031c0e33cc6962e54b7c735a0f0f721fe/internal/config/settings.go#L176)) 

   **Warning is silently dropped in non-verbose mode**

   `utils.Debug` is gated behind `verbose.Load()` (see `internal/utils/debug.go` line 40: `if !verbose.Load() { return }`). In the default, non-verbose startup, this call is a no-op — the corruption warning is never emitted, and the user gets silent fallback to defaults with no indication that their `settings.json` is damaged.

   The PR description explicitly states the intent is to "log warning with file path and error", but `utils.Debug` is a debug-level facility, not a warning channel. A user running without verbose mode will never see or know their settings were silently discarded.

   Consider using `log.Printf` from the standard library (or however the rest of the application surfaces user-facing warnings) so the message is always visible:

   

   This would require adding `"log"` to the import block (which the PR description actually mentions, suggesting this may have been the original intent).

3. `internal/config/settings.go`, line 176 ([link](https://github.com/surge-downloader/surge/blob/a83d26449218f317728cbff80f920afddfe4d3f7/internal/config/settings.go#L176)) 

   **Warning silently discarded — `utils.Debug` is always a no-op here**

   `utils.Debug` drops the message when either (a) verbose mode is off (the default) or (b) `logsDir` has not yet been configured. In the main startup path, `LoadSettings()` is called from `PersistentPreRun` (line 417 of `cmd/root.go`), while `utils.ConfigureDebug(logsDir)` is called later inside `initializeRootLocalRuntime()` → `mustInitializeGlobalState()` which runs in `Run` (line 356 of `cmd/root.go`).

   Because Cobra always executes `PersistentPreRun` before `Run`, `logsDir` is never set at the point `LoadSettings()` fires, so `utils.Debug` returns early at the `val == nil` check in `debug.go:46` every single time — even when `-v` is passed. The user who has a corrupt `settings.json` will silently get defaults with zero indication of what happened.

   The PR description says "log warning", but the chosen call is a debug-file writer that is intentionally invisible in normal operation.

   Use the standard library `log` package (already available in the Go toolchain) or write directly to `os.Stderr` so the warning is always emitted:

   ```go
   import (
       "encoding/json"
       "fmt"
       "os"
       "path/filepath"
       "time"
   )

   // ...

   if err := json.Unmarshal(data, settings); err != nil {
       fmt.Fprintf(os.Stderr, "Warning: corrupt settings file %s: %v — using defaults\n", path, err)
       return DefaultSettings(), nil
   }
   ```

   Or with the `log` package:

   ```go
   log.Printf("Warning: corrupt settings file %s: %v — using defaults", path, err)
   ```

   Either approach guarantees the user sees the warning regardless of verbose flags or log-directory initialization order.

4. `internal/config/settings.go`, line 176 ([link](https://github.com/surge-downloader/surge/blob/34da13edb1e93733a360634fd13eb69bcc389195/internal/config/settings.go#L176)) 

   **Warning silently dropped in non-verbose mode**

   `utils.Debug` is gated behind a `verbose.Load()` flag (see `internal/utils/debug.go` line 40–42). When verbose mode is not enabled — which is the normal production scenario — this call returns immediately without writing anything, so the warning about a corrupt `settings.json` is completely invisible to the user and to any log file.

   This defeats the stated goal of the fix: "log warning with file path and error." A user whose settings file silently falls back to defaults has no way to diagnose the data loss.

   The fix should use an unconditional output path, e.g. `log.Printf` (stdlib) or `fmt.Fprintf(os.Stderr, ...)`:

   ```go
   log.Printf("warning: corrupt settings file %s: %v — using defaults", path, err)
   return DefaultSettings(), nil
   ```

   This also aligns with the PR description, which mentions adding a `log` import, but the actual implementation imported `utils` instead.

5. `internal/config/settings.go`, line 176 ([link](https://github.com/surge-downloader/surge/blob/b1195e2333e3f80ae1e0f56343e033ff8c418dbb/internal/config/settings.go#L176)) 

   **Warning is silently dropped in normal usage**

   `utils.Debug()` is gated behind the `verbose` flag (see `debug.go` line 41: `if !verbose.Load() { return }`). When the app runs without verbose mode enabled — which is the default for end-users — this warning is a no-op. A user whose `settings.json` is corrupt will silently start with defaults and never know their configuration was reset.

   The PR's stated goal is to "log warning with file path and error", but that warning will never appear in standard usage. Consider using the standard `log` package (or `fmt.Fprintf(os.Stderr, ...)`) so the warning is always visible regardless of the verbose flag:

   

   This would also require adding `"log"` to the imports (matching the PR description, which mentions adding a `log` import but the actual diff imports `utils` instead).

</details>

`````markdown
Path: internal/config/settings.go
Line: 176

Comment:
**Warning is silently dropped in normal usage**

`utils.Debug()` is gated behind the `verbose` flag (see `debug.go` line 41: `if !verbose.Load() { return }`). When the app runs without verbose mode enabled — which is the default for end-users — this warning is a no-op. A user whose `settings.json` is corrupt will silently start with defaults and never know their configuration was reset.

The PR's stated goal is to "log warning with file path and error", but that warning will never appear in standard usage. Consider using the standard `log` package (or `fmt.Fprintf(os.Stderr, ...)`) so the warning is always visible regardless of the verbose flag:

```suggestion
		log.Printf("Warning: corrupt settings file %s: %v — using defaults", path, err)
```

This would also require adding `"log"` to the imports (matching the PR description, which mentions adding a `log` import but the actual diff imports `utils` instead).

---

Path: internal/config/settings_test.go
Line: 249-344

Comment:
**Non-English comments in an English codebase**

The three new test functions use Portuguese section labels (`// Cenário`, `// Ação`, `// Validação`) while the entire rest of the test file uses English. This creates an inconsistency that may confuse contributors unfamiliar with Portuguese.

Consider using the English equivalents (`// Arrange`, `// Act`, `// Assert`) to match the rest of the codebase's conventions.

`````

</details>

<sub>Reviews (5): Last reviewed commit: ["test(config): adicionar testes para fall..."](https://github.com/surge-downloader/surge/commit/b1195e2333e3f80ae1e0f56343e033ff8c418dbb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27042535)</sub>

### DIFF
--- a/internal/config/settings_test.go
+++ b/internal/config/settings_test.go
@@ -246,6 +246,99 @@ func TestLoadSettings_CorruptedJSON(t *testing.T) {
 	}
 }
 
+func TestLoadSettings_CorruptedJSON_FallsBackToDefaults(t *testing.T) {
+	// Cenário
+	tmpDir := t.TempDir()
+	t.Setenv("XDG_CONFIG_HOME", tmpDir)
+
+	surgeDir := filepath.Join(tmpDir, "surge")
+	if err := os.MkdirAll(surgeDir, 0o755); err != nil {
+		t.Fatalf("Failed to create surge dir: %v", err)
+	}
+	corruptPath := filepath.Join(surgeDir, "settings.json")
+	if err := os.WriteFile(corruptPath, []byte("{not valid json!!!"), 0o644); err != nil {
+		t.Fatalf("Failed to write corrupt settings: %v", err)
+	}
+
+	// Ação
+	settings, err := LoadSettings()
+
+	// Validação
+	if err != nil {
+		t.Fatalf("LoadSettings should not return error for corrupt JSON, got: %v", err)
+	}
+	if settings == nil {
+		t.Fatal("LoadSettings should return defaults, got nil")
+	}
+
+	defaults := DefaultSettings()
+	if settings.Network.MaxConnectionsPerHost != defaults.Network.MaxConnectionsPerHost {
+		t.Errorf("Expected default MaxConnectionsPerHost %d, got %d",
+			defaults.Network.MaxConnectionsPerHost, settings.Network.MaxConnectionsPerHost)
+	}
+	if settings.Performance.MaxTaskRetries != defaults.Performance.MaxTaskRetries {
+		t.Errorf("Expected default MaxTaskRetries %d, got %d",
+			defaults.Performance.MaxTaskRetries, settings.Performance.MaxTaskRetries)
+	}
+}
+
+func TestLoadSettings_TruncatedJSON_FallsBackToDefaults(t *testing.T) {
+	// Cenário
+	tmpDir := t.TempDir()
+	t.Setenv("XDG_CONFIG_HOME", tmpDir)
+
+	surgeDir := filepath.Join(tmpDir, "surge")
+	if err := os.MkdirAll(surgeDir, 0o755); err != nil {
+		t.Fatalf("Failed to create surge dir: %v", err)
+	}
+	// Simula crash durante SaveSettings — arquivo truncado
+	truncated := `{"general": {"default_download_dir": "/home/user/Downloads", "warn_on_duplicate": tr`
+	corruptPath := filepath.Join(surgeDir, "settings.json")
+	if err := os.WriteFile(corruptPath, []byte(truncated), 0o644); err != nil {
+		t.Fatalf("Failed to write truncated settings: %v", err)
+	}
+
+	// Ação
+	settings, err := LoadSettings()
+
+	// Validação
+	if err != nil {
+		t.Fatalf("LoadSettings should not return error for truncated JSON, got: %v", err)
+	}
+	if settings == nil {
+		t.Fatal("LoadSettings should return defaults, got nil")
+	}
+	if settings.Network.MaxConnectionsPerHost != DefaultSettings().Network.MaxConnectionsPerHost {
+		t.Error("Expected default settings after truncated JSON")
+	}
+}
+
+func TestLoadSettings_EmptyFile_FallsBackToDefaults(t *testing.T) {
+	// Cenário
+	tmpDir := t.TempDir()
+	t.Setenv("XDG_CONFIG_HOME", tmpDir)
+
+	surgeDir := filepath.Join(tmpDir, "surge")
+	if err := os.MkdirAll(surgeDir, 0o755); err != nil {
+		t.Fatalf("Failed to create surge dir: %v", err)
+	}
+	emptyPath := filepath.Join(surgeDir, "settings.json")
+	if err := os.WriteFile(emptyPath, []byte(""), 0o644); err != nil {
+		t.Fatalf("Failed to write empty settings: %v", err)
+	}
+
+	// Ação
+	settings, err := LoadSettings()
+
+	// Validação
+	if err != nil {
+		t.Fatalf("LoadSettings should not return error for empty file, got: %v", err)
+	}
+	if settings == nil {
+		t.Fatal("LoadSettings should return defaults, got nil")
+	}
+}
+
 func TestLoadSettings_PartialJSON(t *testing.T) {
 	// Test that missing fields get filled with defaults
 	partial := `{


LuisMIguelFurlanettoSousa:
Thanks for the review! I've pushed a fix replacing `log.Printf` with `utils.Debug` to match the codebase conventions.

Regarding the Activity Log message: since `LoadSettings()` lives in the `config` package (which doesn't have access to the TUI's `addLogEntry()`), showing a message there would require the caller (in the TUI layer) to check for a corruption signal and display it. I can add that if you'd like — perhaps `LoadSettings()` could return a second value indicating corruption, and the TUI startup code could call `addLogEntry()` with a warning. Let me know if you'd like me to take that approach!

LuisMIguelFurlanettoSousa:
@SuperCoolPencil Done! I've added three tests covering the corrupt settings fallback:

- `TestLoadSettings_CorruptedJSON_FallsBackToDefaults` — invalid JSON → returns defaults without error
- `TestLoadSettings_TruncatedJSON_FallsBackToDefaults` — simulates crash mid-write (truncated file) → defaults
- `TestLoadSettings_EmptyFile_FallsBackToDefaults` — empty file → defaults

All tests pass. Also already replaced `log.Printf` with `utils.Debug` as you requested.

PR #323, opened 2026-04-05
feat: add post-download actions (on-complete and on-error shell hooks)

mvanhorn:
## Summary

Add configurable shell commands that run after a download completes or fails. Commands support template variables for flexible automation.

## Why this matters

aria2 has `--on-download-complete` hooks. wget has `--execute` for post-processing. IDM and JDownloader have post-download actions. Surge currently has no way to automate what happens after a download finishes.

With Surge's daemon architecture, this is especially useful for headless servers and Raspberry Pi setups where users want downloads to auto-extract, move to media directories, or trigger notifications.

## Changes

- Add `PostDownloadActions` config struct with `on_complete_command` and `on_error_command` fields to `settings.json`
- Add `RunPostActions()` in `internal/processing/post_actions.go` with template variable substitution (`{filename}`, `{filepath}`, `{size}`, `{speed}`, `{duration}`, `{id}`, `{error}`)
- Wire into `StartEventWorker` for both `DownloadCompleteMsg` and `DownloadErrorMsg` events
- Actions run in goroutines so they never block the download lifecycle
- Add "Post-Download" settings tab in the TUI
- Cross-platform: uses `sh -c` on Unix, `cmd /C` on Windows

## Testing

All tests pass including new tests for template expansion and command execution:

Tests

```
go test ./... -count=1   # All 18 packages pass
```

This contribution was developed with AI assistance (Codex + Claude Code).

<h3>Greptile Summary</h3>

This PR adds configurable shell hooks (`on_complete_command` / `on_error_command`) that fire after a download completes or fails, with template variable substitution and proper shell escaping. Previous concerns about injection, incorrect event routing, and missing `destPath` fallback have all been addressed.

- **P1**: The `DownloadCompleteMsg` finalization-error path (`finalizeCompletedFile` failure) hits `break` before either post-action block is reached, so `OnErrorCommand` never fires when the downloaded file can't be renamed/moved — even though the download is recorded as `\"error\"` and a notification fires.

Safe to merge after fixing the finalization-error path missing the OnErrorCommand hook.

All prior review concerns (injection, routing, destPath, redundant GetSettings) are resolved. One new P1 remains: the finalization-error branch in DownloadCompleteMsg breaks out before RunPostActions, silently skipping the error hook for a named failure scenario that already fires a notification.

internal/processing/events.go — the finalizeCompletedFile error branch (lines 227–253) needs an OnErrorCommand hook call before the break.

| Filename | Overview |
|----------|----------|
| internal/processing/events.go | Wires RunPostActions into both DownloadCompleteMsg and DownloadErrorMsg; finalization-error path breaks before firing OnErrorCommand hook (P1). |
| internal/processing/post_actions.go | New file implementing shell hooks with proper shellEscape, template expansion, 30s timeout, and explicit isError flag; previous injection and routing bugs are resolved. |
| internal/config/settings.go | Adds PostDownloadActions struct and inserts it into the Settings struct between General and Network; clean addition with correct JSON tags and ui_desc variable documentation. |
| internal/processing/post_actions_test.go | Good coverage of template expansion including injection-attempt edge cases; does not cover the finalization-error path gap in events.go. |
| internal/config/settings_test.go | Minor update: bumps expected category count from 5 to 6 to account for the new Post-Download section. |
| internal/tui/view_settings.go | Adds Post-Download reset-to-default cases; follows existing patterns in the switch statement correctly. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant EW as EventWorker
    participant PA as RunPostActions
    participant SH as sh/cmd

    EW->>EW: DownloadCompleteMsg received
    EW->>EW: finalizeCompletedFile()
    alt finalization fails
        EW->>EW: mark status=error, notify
        Note over EW: break — hooks never reached
    else finalization succeeds
        EW->>EW: mark status=completed, notify
        EW->>+PA: go RunPostActions(OnCompleteCommand, isError=false)
        PA->>PA: expandTemplate (shellEscape vars)
        PA->>+SH: sh -c expanded (30s timeout)
        SH-->>-PA: output / exit code
        PA-->>-EW: goroutine exits
    end

    EW->>EW: DownloadErrorMsg received
    EW->>EW: mark status=error, notify
    EW->>+PA: go RunPostActions(OnErrorCommand, isError=true)
    PA->>PA: expandTemplate (shellEscape vars)
    PA->>+SH: sh -c expanded (30s timeout)
    SH-->>-PA: output / exit code
    PA-->>-EW: goroutine exits
```

<details open><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `internal/tui/view_settings.go`, line 247-283 ([link](https://github.com/surgedm/surge/blob/f5f21de9d3ccf202bdb7b23596034e8a35a3277f/internal/tui/view_settings.go#L247-L283)) 

   <a href="#">P1</a> **"Post-Download" tab is silently non-functional in the TUI**

   `getSettingsValues` has no `"Post-Download"` case, so the tab always renders empty values. `setSettingValue` (line 298) and `resetSettingToDefault` (line 631) also lack this case, meaning every edit the user types is silently discarded — the feature is only configurable by hand-editing `settings.json`.

   Three switch blocks all need a `"Post-Download"` arm. For example in `getSettingsValues`:

   ```go
   case "Post-Download":
       values["on_complete_command"] = m.Settings.General.PostDownload.OnCompleteCommand
       values["on_error_command"] = m.Settings.General.PostDownload.OnErrorCommand
   ```

   And in `setSettingValue`, route to a new helper:

   ```go
   case "Post-Download":
       if key == "on_complete_command" {
           m.Settings.General.PostDownload.OnCompleteCommand = value
       } else if key == "on_error_command" {
           m.Settings.General.PostDownload.OnErrorCommand = value
       }
   ```

   And in `resetSettingToDefault`:

   ```go
   case "Post-Download":
       switch key {
       case "on_complete_command":
           m.Settings.General.PostDownload.OnCompleteCommand = defaults.General.PostDownload.OnCompleteCommand
       case "on_error_command":
           m.Settings.General.PostDownload.OnErrorCommand = defaults.General.PostDownload.OnErrorCommand
       }
   ```

   **Rule Used:** What: Enforce separation of concerns by keeping bu... ([source](https://app.greptile.com/review/custom-context?memory=c1138776-52c4-4b49-81a0-1cd68fe194dc))
   
   `````markdown
   Path: internal/tui/view_settings.go
   Line: 247-283
   
   Comment:
   **"Post-Download" tab is silently non-functional in the TUI**
   
   `getSettingsValues` has no `"Post-Download"` case, so the tab always renders empty values. `setSettingValue` (line 298) and `resetSettingToDefault` (line 631) also lack this case, meaning every edit the user types is silently discarded — the feature is only configurable by hand-editing `settings.json`.
   
   Three switch blocks all need a `"Post-Download"` arm. For example in `getSettingsValues`:
   
   ```go
   case "Post-Download":
       values["on_complete_command"] = m.Settings.General.PostDownload.OnCompleteCommand
       values["on_error_command"] = m.Settings.General.PostDownload.OnErrorCommand
   ```
   
   And in `setSettingValue`, route to a new helper:
   
   ```go
   case "Post-Download":
       if key == "on_complete_command" {
           m.Settings.General.PostDownload.OnCompleteCommand = value
       } else if key == "on_error_command" {
           m.Settings.General.PostDownload.OnErrorCommand = value
       }
   ```
   
   And in `resetSettingToDefault`:
   
   ```go
   case "Post-Download":
       switch key {
       case "on_complete_command":
           m.Settings.General.PostDownload.OnCompleteCommand = defaults.General.PostDownload.OnCompleteCommand
       case "on_error_command":
           m.Settings.General.PostDownload.OnErrorCommand = defaults.General.PostDownload.OnErrorCommand
       }
   ```
   
   **Rule Used:** What: Enforce separation of concerns by keeping bu... ([source](https://app.greptile.com/review/custom-context?memory=c1138776-52c4-4b49-81a0-1cd68fe194dc))
   
   `````
   </details>

2. `internal/processing/events.go`, line 227-253 ([link](https://github.com/surgedm/surge/blob/67fc35f27324a6700b27065839331237397a9970/internal/processing/events.go#L227-L253)) 

   <a href="#">P1</a> **Finalization error silently skips the `OnErrorCommand` hook**

   When `finalizeCompletedFile` fails (e.g., the `.surge` rename fails because of a permissions error or a cross-device move), the code marks the download as `"error"` and fires a system notification, then hits `break` — which exits the `DownloadCompleteMsg` case entirely. Neither the `OnCompleteCommand` block (lines 292–303) nor the `OnErrorCommand` block in the `DownloadErrorMsg` case is ever reached. Users who configure `on_error_command` to trigger alerting or cleanup will silently get nothing for this failure path.

   ```go
   if err := finalizeCompletedFile(destPath); err != nil {
       // ... existing error-state persistence and notification ...
       if settings != nil && settings.PostDownload.OnErrorCommand != "" {
           go RunPostActions(settings.PostDownload, PostActionContext{
               Filename: filename,
               FilePath: destPath,
               ID:       m.DownloadID,
               Error:    err.Error(),
           }, true)
       }
       break
   }
   ```

   `````markdown
   Path: internal/processing/events.go
   Line: 227-253

   Comment:
   **Finalization error silently skips the `OnErrorCommand` hook**

   When `finalizeCompletedFile` fails (e.g., the `.surge` rename fails because of a permissions error or a cross-device move), the code marks the download as `"error"` and fires a system notification, then hits `break` — which exits the `DownloadCompleteMsg` case entirely. Neither the `OnCompleteCommand` block (lines 292–303) nor the `OnErrorCommand` block in the `DownloadErrorMsg` case is ever reached. Users who configure `on_error_command` to trigger alerting or cleanup will silently get nothing for this failure path.

   ```go
   if err := finalizeCompletedFile(destPath); err != nil {
       // ... existing error-state persistence and notification ...
       if settings != nil && settings.PostDownload.OnErrorCommand != "" {
           go RunPostActions(settings.PostDownload, PostActionContext{
               Filename: filename,
               FilePath: destPath,
               ID:       m.DownloadID,
               Error:    err.Error(),
           }, true)
       }
       break
   }
   ```

   `````
   </details>

</details>

`````markdown
Path: internal/processing/events.go
Line: 227-253

Comment:
**Finalization error silently skips the `OnErrorCommand` hook**

When `finalizeCompletedFile` fails (e.g., the `.surge` rename fails because of a permissions error or a cross-device move), the code marks the download as `"error"` and fires a system notification, then hits `break` — which exits the `DownloadCompleteMsg` case entirely. Neither the `OnCompleteCommand` block (lines 292–303) nor the `OnErrorCommand` block in the `DownloadErrorMsg` case is ever reached. Users who configure `on_error_command` to trigger alerting or cleanup will silently get nothing for this failure path.

```go
if err := finalizeCompletedFile(destPath); err != nil {
    // ... existing error-state persistence and notification ...
    if settings != nil && settings.PostDownload.OnErrorCommand != "" {
        go RunPostActions(settings.PostDownload, PostActionContext{
            Filename: filename,
            FilePath: destPath,
            ID:       m.DownloadID,
            Error:    err.Error(),
        }, true)
    }
    break
}
```

`````

</details>

<sub>Reviews (9): Last reviewed commit: ["Merge branch &#39;upstream/main&#39; into feat/p..."](https://github.com/surgedm/surge/commit/67fc35f27324a6700b27065839331237397a9970) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27384520)</sub>


mvanhorn:
Fixed the shell injection in `expandTemplate` - template variables like `{filename}` are now shell-escaped before substitution (single-quoted on Unix, double-quoted on Windows). Pushed in 7de5a0d.

The P1 (empty error routing to OnComplete) is a theoretical edge case since error messages are always non-empty in practice. The P2 about redundant `GetSettings()` is incorrect - the variable is used at both line 275 and 292.

mvanhorn:
Fixed the Windows CI failures. `TestExpandTemplate` had single-quoted expected values hardcoded, but `shellEscape` uses double quotes on Windows per the runtime check. Rebuilt the expected strings by calling `shellEscape` directly so the test validates template substitution without pinning a quoting style. Passing locally on Unix and should cover Windows now (57e2b7a).

mvanhorn:
Addressed the greptile findings:

- **P0 shell injection**: Already handled — `expandTemplate` passes all string variables through `shellEscape()` (single-quote wrapping on Unix, double-quote on Windows with internal quote escaping). Numeric fields use `fmt.Sprintf` with `%d`/`%.2f` so they can't inject.
- **P1 empty error routing**: Already handled — line 340 checks `m.Err != nil` before running the error post-action, so a nil-error `DownloadErrorMsg` correctly skips it.
- **P1 destPath in error hook**: Fixed in e3a9798 — the error post-action was using `m.DestPath` instead of the resolved `destPath` (which falls back to `existing.DestPath` when the message field is empty).
- **P2 redundant GetSettings / restating comments**: Each case block calls `GetSettings()` once; removed the comment that restated the function call.

mvanhorn:
Pushed 071d26e addressing the latest greptile P1s and P2s.

P1 on-error hook silent when `m.Err == nil` (events.go:340): the notification block already defaults to "Download failed" on nil Err, so users configuring `on_error_command` expect their hook to fire on the same event. Dropped the `m.Err != nil` gate, still pass an empty error string when Err is nil.

P1 Post-Download TUI tab non-functional (view_settings.go): added `Post-Download` cases in `getSettingsValues`, `setSettingValue`, and `resetSettingToDefault` so the tab reads, writes, and resets `on_complete_command` / `on_error_command`.

P2 timeout (post_actions.go): wrapped shell exec in `context.WithTimeout(30s)` so hung user commands don't leak goroutines on headless deployments.

P2 comment (events.go:293): replaced `// Run post-download actions` with the why, the goroutine keeps slow shell commands off the serialized event worker.

Incidental: fixed pre-existing `TestExpandTemplate/all_vars` that was broken on this branch (e5adac8 wrapped `{duration}` in `shellEscape()` but didn't update the expected value). Also added shellEscape edge-case tests covering semicolons, dollar signs, backticks, and embedded quotes.

`go test ./... -count=1` passes on all 18 packages. `go vet ./...` clean.

mvanhorn:
Addressed the remaining P2s in 1429f84: added `OnCompleteCommand != ""` guard to the complete handler (matches the error path, avoids a no-op goroutine per download), and rewrote the error hook comment to clarify it fires independently of notifications. `go test ./internal/processing/...` passes.

PR #362, opened 2026-04-12
feat: Add import and export functionality in surge

junaid2005p:
<h3>Greptile Summary</h3>

This PR adds import/export (backup/restore) functionality to Surge via new CLI commands (`export`, `import`), a TUI Data Transfer panel, HTTP API endpoints (`/data/export`, `/data/import/preview`, `/data/import/apply`), and a `backup` package with a staged two-phase preview→apply workflow.

The three major security issues raised in the previous review (predictable session ID, missing upload size limit, relative path traversal) are all addressed in this revision.

Safe to merge for most paths; Replace-mode import has a known data-integrity ordering issue from the prior review that remains unresolved.

All three prior security findings (path traversal, upload limit, predictable session ID) are fixed. New findings are P2 only (memory efficiency in verifyManifestFiles, silent marshal error, missing LeavePaused TUI toggle). Score stays at 4 because the non-atomic Replace ordering (clear DB → save settings → restore) is still present and represents a real data-loss risk for Replace-mode imports that abort midway.

internal/backup/backup.go — ApplyImport ordering in Replace mode; verifyManifestFiles memory usage

| Filename | Overview |
|----------|----------|
| internal/backup/backup.go | Core export/import logic; verifyManifestFiles reads entire zip entries into memory via io.ReadAll for SHA-256, which is wasteful for large partial files; settings are saved before downloads/logs are fully restored in Replace mode (flagged in prior review, still present) |
| cmd/transfer_api.go | HTTP endpoints for export/import with size cap, crypto-random session IDs, and staged session store; json.Marshal error is silently discarded in the export manifest header |
| internal/backup/path.go | Path rebase/sanitization logic with containment checks via pathWithinRoot; previously flagged path traversal is fixed; test coverage added in path_test.go |
| internal/core/transfer.go | Clean TransferService interface with local and remote implementations; remote client correctly ignores src for apply calls; 5-minute HTTP timeout is a reasonable default |
| internal/tui/update_transfer.go | TUI export/import command handlers; LeavePaused option not surfaced despite being supported by ExportOptions; all async commands use context.Background() so they can't be cancelled if TUI exits |
| cmd/transfer_utils.go | CLI helpers for resolve, export, preview, and apply bundle; path cleaning applied consistently; session ID plumbed correctly for remote vs local paths |
| cmd/export.go | New export CLI command; all flags wired correctly; JSON output via shared printJSON helper |
| cmd/import.go | New import CLI command with preview-then-apply flow; SessionID plumbed correctly for remote sessions; options copied correctly from preview to apply |
| internal/backup/backup_test.go | Good test coverage for partial restore, log restore, replace/merge modes, and path escape rejection |

</details>

`````markdown
Path: internal/backup/backup.go
Line: 732-748

Comment:
**`verifyManifestFiles` loads full zip entries into memory**

Each entry is read with `io.ReadAll` before being hashed, so a bundle with a large partial file (up to the 512 MB cap) allocates that entire file in RAM. A streaming approach is more efficient:

```go
func verifyManifestFiles(reader *zip.Reader, files []ManifestFile) error {
    for _, mf := range files {
        var found *zip.File
        for _, f := range reader.File {
            if filepath.ToSlash(f.Name) == filepath.ToSlash(mf.Path) {
                found = f
                break
            }
        }
        if found == nil {
            return fmt.Errorf("bundle entry %s not found", mf.Path)
        }
        rc, err := found.Open()
        if err != nil {
            return err
        }
        h := sha256.New()
        size, err := io.Copy(h, rc)
        _ = rc.Close()
        if err != nil {
            return err
        }
        if hex.EncodeToString(h.Sum(nil)) != mf.SHA256 {
            return fmt.Errorf("checksum mismatch for %s", mf.Path)
        }
        if size != mf.Size {
            return fmt.Errorf("size mismatch for %s", mf.Path)
        }
    }
    return nil
}
```

---

Path: cmd/transfer_api.go
Line: 87

Comment:
**Silent marshal error discards manifest header**

If `json.Marshal(manifest)` fails, `manifestBytes` is `nil`, the `X-Surge-Manifest` header is sent as an empty string, and the remote client silently returns `&backup.Manifest{}`. Log or handle the error so callers aren't left with a structureless manifest on unexpected failures.

```suggestion
		manifestBytes, err := json.Marshal(manifest)
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}
```

---

Path: internal/tui/update_transfer.go
Line: 28-31

Comment:
**`LeavePaused` not exposed in TUI export**

`ExportOptions.LeavePaused` controls whether paused downloads are automatically resumed after export, and is available in the CLI via `--leave-paused`. The TUI omits this toggle, so interactive users have no way to keep downloads paused. Consider adding a `m.transferLeavePaused` state field alongside the other toggles and wiring it here.

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: fixed COdeQL issues"](https://github.com/surgedm/surge/commit/71b7995e186ac574be62c8d406331b18ec131262) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28117134)</sub>

### DIFF
--- a/cmd/connect.go
+++ b/cmd/connect.go
@@ -78,7 +78,7 @@ func connectAndRunTUI(cmd *cobra.Command, target string) error {
 		}
 	}
 
-	m := newRemoteRootModel(port, service, serverHost)
+	m := newRemoteRootModel(port, service, serverHost, baseURL, token)
 
 	p := tea.NewProgram(m)
 	go func() {
@@ -93,8 +93,9 @@ func connectAndRunTUI(cmd *cobra.Command, target string) error {
 	return nil
 }
 
-func newRemoteRootModel(port int, service core.DownloadService, serverHost string) tui.RootModel {
+func newRemoteRootModel(port int, service core.DownloadService, serverHost string, baseURL string, token string) tui.RootModel {
 	m := tui.InitialRootModel(port, Version, service, nil, false)
+	m.Transfer = core.NewRemoteTransferService(baseURL, token)
 	m.ServerHost = serverHost
 	m.IsRemote = true
 	return m
--- a/cmd/connect_test.go
+++ b/cmd/connect_test.go
@@ -65,7 +65,7 @@ func (f *fakeRemoteDownloadService) GetStatus(id string) (*types.DownloadStatus,
 func (f *fakeRemoteDownloadService) Shutdown() error { return nil }
 
 func TestNewRemoteRootModel_UsesNilOrchestrator(t *testing.T) {
-	m := newRemoteRootModel(1700, nil, "example.com")
+	m := newRemoteRootModel(1700, nil, "example.com", "https://example.com:1700", "token")
 
 	if m.Orchestrator != nil {
 		t.Fatal("expected remote root model to use nil orchestrator")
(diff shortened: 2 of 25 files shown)


SuperCoolPencil:
stale

PR #375, opened 2026-04-16
Fix WSL download path handling

### LINKED ISSUE #365: Wrong download directory path when using WSL
**Describe the bug**
I configured surge docker-compose's file to mount the download directory on `/mnt/c/Users/me/Downloads` so that I could easily find my files downloaded through Surge. It happen to be working almost as intended : Surge saves the file in `/mnt/c/Users/me/Downloads/C:/Users/me/Downloads/`.

**To Reproduce**
1. Download compose.yml file as explained in README.md
2. Set volumes to:
```yaml 
  volumes:
      - /mnt/c/Users/me/Downloads:/downloads
      - ./surge-config:/root/.local/state/surge
```
4. Boot up the container / setup the token with the extension.
5. Download a file : the file gets put in `/mnt/c/Users/me/Downloads/C:/Users/me/Downloads/` instead of `/mnt/c/Users/me/Downloads/`.

**Expected behavior**
The file should be put directly in `/mnt/c/Users/me/Downloads/` (as when you use `~/surge/downloads` for example)
 
**Screenshots**
Image
Image
Image

**Logs**
Surge automatically writes debug log files. (Not with docker ? the log dir remains empty)

**System info:**
- OS: Windows 11 with WSL 2.7.1.0
- Surge Version: latest on docker image (v0.7.5)
- Installed From: Docker images

**Additional context**
- I know the docker image is a bit behind as the current version is v8.0 but I did not see any fix in the changelogs anyway
- I tried to trace down the issue in the source code but without great success for now. I'll keep updating in this issue.

utafrali:
Fixes #365

Windows absolute paths from the browser extension were being joined with the default path as if they were relative subdirectories. Added a check to detect and skip Windows paths. Includes tests for both forward and backslash formats.

<h3>Greptile Summary</h3>

This PR fixes WSL/Linux handling of Windows-style absolute paths sent by the browser extension, preventing them from being incorrectly joined as a relative subdirectory of the default download directory. However, there are three issues that need to be addressed before merge:

- `internal/utils/path_test.go` adds a second `TestIsWindowsAbsPath` on top of an existing one — this is a compile error that blocks the test suite from building.
- The refactored `IsWindowsAbsPath` silently drops UNC-path detection (`\\\\server\\share`), but the pre-existing test at line 82 still asserts UNC paths return `true`, so the existing test will fail once the duplicate is resolved.
- The new `IsWindowsAbsPath` guard added to `resolveOutputDir` is unreachable dead code: `mapClientWindowsPath` (called first) always returns a non-empty value for Windows paths when `relativeToDefaultDir=true`, so execution never reaches the new check.

Not safe to merge — the test package will not compile due to a duplicate function name, and the new implementation has a dead-code guard and drops UNC path support without cleaning up the existing test.

A P0 compile error (duplicate TestIsWindowsAbsPath) prevents the test suite from building entirely. There are also two P1 issues: an existing UNC-path test that would fail once the duplicate is fixed, and a dead-code guard that gives false assurance while the real fix is in mapClientWindowsPath.

internal/utils/path_test.go (duplicate function / compile error) and internal/utils/path.go (UNC support removal) need immediate attention; cmd/root_downloads.go dead-code guard should also be cleaned up.

| Filename | Overview |
|----------|----------|
| internal/utils/path_test.go | Added a new TestIsWindowsAbsPath function that duplicates the existing one — causes a compile error; also the existing function tests UNC paths that the new implementation no longer supports. |
| internal/utils/path.go | Simplified IsWindowsAbsPath removes UNC path support and strings.TrimSpace; also missing trailing newline. UNC removal breaks an existing test case. |
| cmd/root_downloads.go | New IsWindowsAbsPath guard added inside the relativeToDefaultDir branch is unreachable dead code — mapClientWindowsPath already handles all Windows paths before this block can execute. |
| cmd/utils_test.go | Adds TestResolveOutputDir covering WSL forward-slash, backslash, normal subdir, and empty-path cases; tests are correct but exercise mapClientWindowsPath rather than the new dead-code guard. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["resolveOutputDir(reqPath, relativeToDefault, defaultOutputDir, settings)"] --> B["mapClientWindowsPath(reqPath, ...)"]
    B --> C{IsWindowsAbsPath?}
    C -- No --> D["return ''"]
    C -- Yes --> E["MapWindowsPathToDefaultDir(reqPath, baseDir)"]
    E -- ok=true --> F["return mapped path"]
    E -- ok=false --> G{"shouldFallbackUnmappedWindowsPath\n(relativeToDefault || OS≠windows)"}
    G -- false --> H["return ''"]
    G -- true --> I["return filepath.Clean(baseDir)"]
    D --> J{relativeToDefault && reqPath≠''}
    H --> J
    J -- Yes --> K["IsWindowsAbsPath(reqPath)\n⚠️ Dead code — never reached for Windows paths on Linux"]
    K -- true --> L["return baseDir"]
    K -- false --> M["filepath.Join(baseDir, reqPath)"]
    J -- No --> N{outPath == ''}
    N -- Yes --> O["use defaultOutputDir / settings / '.'"]
    N -- No --> P["return outPath"]
    F --> Q["return mapped ✓"]
    I --> Q
```

`````markdown
Path: internal/utils/path_test.go
Line: 9-31

Comment:
**Duplicate function name causes compile error**

This PR adds a new `TestIsWindowsAbsPath` starting at line 9, but the file already contains a `TestIsWindowsAbsPath` at line 76. Go does not allow two functions with the same name in the same package — this is a compile-time error that will prevent the entire test suite from building.

The existing function at line 76 also includes a UNC-path case (`\\server\share\file.zip → true`) that the new simplified `IsWindowsAbsPath` implementation no longer supports, so the two functions are inconsistent. One of them must be removed and the other must reflect the current contract of the function (with or without UNC support).

---

Path: internal/utils/path.go
Line: 22-28

Comment:
**UNC path detection silently dropped**

The original `IsWindowsAbsPath` also recognised UNC paths (e.g. `\\server\share\file.zip`). The refactored version only handles drive-letter paths. The existing test at line 82 of this file asserts `\\server\share\file.zip → true`; with this change that assertion will fail once the duplicate-function compile error is resolved.

If UNC paths are intentionally out of scope for this fix, the corresponding test case must be removed (or updated) and a comment should explain why UNC paths are no longer considered Windows-absolute. If they should still be handled, the old two-character prefix check needs to be restored.

---

Path: cmd/root_downloads.go
Line: 361-363

Comment:
**New guard is unreachable dead code**

This block is inside `if relativeToDefaultDir && reqPath != ""`. For that branch to be reached, `mapClientWindowsPath` (line 347) must have returned `""`. However, when `relativeToDefaultDir == true`, `shouldFallbackUnmappedWindowsPath` always returns `true` regardless of OS, so `mapClientWindowsPath` always returns a non-empty string for any Windows-absolute path (either the mapped result or `filepath.Clean(baseDir)` as the fallback). The new `IsWindowsAbsPath` check can therefore never execute.

The actual protection against joining a Windows path as a subdirectory is already provided by `mapClientWindowsPath`; this extra guard gives a false sense of safety while adding confusion. It should either be removed, or the comment should be on the `mapClientWindowsPath` call where the real work happens.

---

Path: internal/utils/path.go
Line: 79

Comment:
**Missing newline at end of file**

The diff shows `\ No newline at end of file`. Go source files (and most linters/formatters) expect a trailing newline.

```suggestion
}
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["Fix WSL download path handling"](https://github.com/surgedm/surge/commit/cbb7d8735e7cb87ba08b60f50dfb602ab8901c4d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28589064)</sub>

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/cmd/root_downloads.go
+++ b/cmd/root_downloads.go
@@ -356,6 +356,11 @@ func resolveOutputDir(reqPath string, relativeToDefaultDir bool, defaultOutputDi
 		if baseDir == "" {
 			baseDir = "."
 		}
+		// Windows absolute path from browser extension on Linux/Docker host:
+		// treat it as the default dir to avoid joining it as a subdirectory.
+		if utils.IsWindowsAbsPath(reqPath) {
+			return baseDir
+		}
 		outPath = filepath.Join(baseDir, reqPath)
 	} else if outPath == "" {
 		if defaultOutputDir != "" {
--- a/cmd/utils_test.go
+++ b/cmd/utils_test.go
@@ -5,6 +5,8 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/SurgeDM/Surge/internal/config"
 )
 
 func TestResolveClientOutputPath(t *testing.T) {
@@ -100,6 +102,58 @@ func TestResolveClientOutputPath(t *testing.T) {
 				if err != nil || strings.HasPrefix(rel, "..") {
 					t.Errorf("resolveClientOutputPath(%q) = %q, want prefix %q", tt.outputDir, got, tt.wantPrefix)
 				}
+			}
+		})
+	}
+}
+
+func TestResolveOutputDir(t *testing.T) {
+	settings := config.DefaultSettings()
+	settings.General.DefaultDownloadDir = "/downloads"
+
+	tests := []struct {
+		name                string
+		reqPath             string
+		relativeToDefault   bool
+		defaultOutputDir    string
+		want                string
+	}{
+		{
+			name:              "WSL Windows path with relativeToDefaultDir",
+			reqPath:           "C:/Users/me/Downloads",
+			relativeToDefault: true,
+			defaultOutputDir:  "/downloads",
+			want:              "/downloads",
+		},
+		{
+			name:              "WSL Windows path backslash with relativeToDefaultDir",
+			reqPath:           `C:\Users\me\Downloads`,
+			relativeToDefault: true,
+			defaultOutputDir:  "/downloads",
+			want:              "/downloads",
+		},
+		{
+			name:              "normal relative subdir",
+			reqPath:           "subdir",
+			relativeToDefault: true,
+			defaultOutputDir:  "/downloads",
+			want:              "/downloads/subdir",
+		},
+		{
+			name:              "empty path uses default",
+			reqPath:           "",
+			relativeToDefault: false,
+			defaultOutputDir:  "/downloads",
+			want:              "/downloads",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := resolveOutputDir(tt.reqPath, tt.relativeToDefault, tt.defaultOutputDir, settings)
+			if got != tt.want {
+				t.Errorf("resolveOutputDir(%q, %v, %q) = %q, want %q",
+					tt.reqPath, tt.relativeToDefault, tt.defaultOutputDir, got, tt.want)
 			}
 		})
 	}
--- a/internal/utils/path.go
+++ b/internal/utils/path.go
@@ -18,23 +18,13 @@ func EnsureAbsPath(path string) string {
 	return path
 }
 
-// IsWindowsAbsPath reports whether p looks like a Windows absolute path even
-// when running on a non-Windows host (for example inside Docker on Linux).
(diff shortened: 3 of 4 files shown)


SuperCoolPencil:
#365 was already fixed by #366

PR #427, opened 2026-05-03
fix: gate service mode behind internal subcommand

tetra-fox:
`Execute()` routes every invocation through `kardianos/service`. Its `Getppid() != 1` check misfires in nested PID namespaces (nix sandboxes, `unshare --pid`, CI), hanging `surge completion zsh`, `--version`, etc. forever. #424 patched this for Android. This removes the underlying check for every platform.

The installer now points the unit's `ExecStart` at a hidden `service __run` subcommand whose only job is to call `RunService()`. Every other invocation runs cobra directly and exits when done.

### Reproduction

```sh
unshare -r --pid --fork --mount-proc sh -c "surge --version >/dev/null"
# main: hangs indefinitely
# fix: exits cleanly
```

### Notes

* Linux/macOS/BSD pre-existing service installs keep working because `surge server start` handles its own SIGTERM/context.
* Windows SCM users should re-run `surge service install` to pick up the new entry point (a `svc.IsWindowsService()` fallback could fix this in a follow-up).
* Termux unchanged.

<h3>Greptile Summary</h3>

This PR fixes a hang in `kardianos/service`'s `Getppid() != 1` detection misfiring inside nested PID namespaces by moving all normal invocations onto a direct `rootCmd.Execute()` path. A new hidden `service __run` subcommand becomes the sole entry point for the service manager, and `p.Start()` explicitly overrides args to `server start` before re-entering cobra to avoid re-dispatch recursion.

Safe to merge; the fix is targeted and the two previously-identified issues (recursion on error path and SetArgs mutation) are both correctly addressed and documented.

All three files have focused, correct changes. The infinite-recursion bug in the error path is fixed by returning the error directly. The permanent SetArgs mutation is intentional and explained with a clear comment. No new P0/P1 issues found.

| Filename | Overview |
|----------|----------|
| cmd/root.go | Single-line change: `Execute()` now calls `rootCmd.Execute()` directly instead of routing through `RunService()`, breaking the service-detection code path for all normal invocations. |
| cmd/runservice_std.go | Simplified to remove the `Interactive()` fallback and the recursive `rootCmd.Execute()` error fallback; now returns `GetService()` errors directly and always calls `s.Run()` on success. |
| cmd/service_kardianos.go | Adds hidden `service __run` subcommand as the new service entry point; updates `serviceConfig.Arguments` from `["server","start"]` to `["service","__run"]`; adds `rootCmd.SetArgs([]string{"server","start"})` before executing to prevent re-dispatch loop, with an explanatory comment. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix: prevent recursion when GetService f..."](https://github.com/surgedm/surge/commit/eea92543c1adb94e690372cf6203c2bffd143c01) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30580442)</sub>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -591,7 +591,7 @@ func performRestart() error {
 
 // Execute adds all child commands to the root command and sets flags appropriately.
 func Execute() error {
-	return RunService()
+	return rootCmd.Execute()
 }
 
 func init() {
--- a/cmd/runservice_std.go
+++ b/cmd/runservice_std.go
@@ -5,7 +5,7 @@ package cmd
 func RunService() error {
 	s, err := GetService()
 	if err != nil {
-		return rootCmd.Execute()
+		return err
 	}
 	return s.Run()
 }
--- a/cmd/runservice_std.go
+++ b/cmd/runservice_std.go
@@ -2,22 +2,10 @@
 
 package cmd
 
-import (
-	"github.com/kardianos/service"
-)
-
-// RunService handles the application execution, checking if it should run as a service.
-// On non-Android platforms, we use kardianos/service's detection to decide whether
-// to run interactively or as a managed service.
 func RunService() error {
 	s, err := GetService()
 	if err != nil {
 		return rootCmd.Execute()
 	}
-
-	if service.Interactive() {
-		return rootCmd.Execute()
-	}
-
 	return s.Run()
 }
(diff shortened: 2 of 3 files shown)


SuperCoolPencil (approved):

SuperCoolPencil:
Thank you so much for contributing :)

🚀

PR #435, opened 2026-05-10
fix: support user-level service and system service fallbacks

### LINKED ISSUE #433: Failed to connect to surge after service start
**Describe the bug**
Cannot connect to surge server after starting surge service.

**To Reproduce**
1. Start surge service:
    ```shell
    # powershell in administrator mode
    surge service install
    surge service start
    surge service status
    # Service is running
    ```
2. Connect to surge server:
    ```shell
    surge connect
    # no local Surge server detected. Start one with 'surge' or 'surge server', or specify a target: surge connect <host:port>
    ```
3. `surge add` also failed:
    ```shell
    surge add https://github.com/SurgeDM/Surge/releases/download/v0.8.5/Surge_0.8.5_windows_amd64.zip
    # surge is not running locally. start it or pass --host (or set SURGE_HOST)
    ```

**Expected behavior**
Connect to surge server successfully.

**Screenshots**
I can find surge service in windows task manager.

Image

**Logs**
Log connect in `C:\Windows\System32\config\systemprofile\AppData\Roaming\surge\logs`:
```txt
[2026-05-08 20:23:02] Startup integrity check: no issues found
[2026-05-08 20:23:02] HTTP server listening on port 1700
```

- OS: Windows 11
- Surge Version: Surge v0.8.5
- Commit: Not built from source
- Installed From: scoop

**Additional context**
I can start surge server with `surge server` and connect to it sucessfully.

SuperCoolPencil:
Resolves #433

This PR introduces a `--user` flag to `surge service` commands to allow running Surge as a user-level service without elevated privileges. 

Additionally, it implements fallbacks so `surge connect` can auto-detect a system-wide service daemon on Windows (LocalSystem) and Linux (root) as long as it has sufficient permissions.

<h3>Greptile Summary</h3>

This PR adds a `--user` flag to `surge service` subcommands so Surge can be installed and managed as a user-level daemon (systemd user instance on Linux, LaunchAgent on macOS) without elevated privileges. It also adds auto-detection fallback logic in the client so that `surge connect` can locate a system-wide service running as root/LocalSystem when no user-level service is found.

- `GetSystemStateDir()` / `GetSystemRuntimeDir()` derive the root/LocalSystem state paths using `user.Lookup(\"root\")` on POSIX and `%SystemRoot%\\System32\\config\\systemprofile` on Windows, matching the paths the daemon actually writes during normal operation.
- `getActiveConnectionDetails()` unifies port and token resolution into a single function with a user-first, system-fallback priority; however, `readActivePort()` and `resolveLocalToken()` each call it independently, so there is a TOCTOU window where the port and token can be sourced from different service instances in a race.
- The `--user` flag is not guarded against Windows, where `kardianos/service` silently ignores the `UserService` option, leaving the service installed as a system service and the user expecting privilege-free installation facing an unexplained access-denied error.

Not safe to merge as-is: the port and token can be sourced from different service instances in the same connection attempt, and the --user flag silently misbehaves on Windows.

The new getActiveConnectionDetails() function correctly pairs port with its corresponding token file, but readActivePort() and resolveLocalToken() each invoke it independently. A service stopping between the two calls can return a port from the user service and a token from the system service (or vice versa), sending the wrong bearer token and producing an opaque auth failure. Separately, --user on Windows passes UserService:true to kardianos/service, which ignores it, so the flag appears to work but installs a system service instead — a user expecting no elevation requirement will hit an unexplained access-denied error.

cmd/utils.go (double invocation of getActiveConnectionDetails) and cmd/service_kardianos.go (missing Windows guard on --user)

| Filename | Overview |
|----------|----------|
| cmd/utils.go | Introduces getActiveConnectionDetails() for coupled port+token resolution, but readActivePort() and resolveLocalToken() each call it independently, creating a TOCTOU window where port and token can be sourced from different service instances. |
| cmd/service_kardianos.go | Adds --user PersistentFlag and merges UserService option safely into service config; flag has no Windows guard, so it silently no-ops on Windows and can mislead users expecting privilege-free installation. |
| internal/config/paths.go | Adds GetSystemStateDir() and GetSystemRuntimeDir() that mirror the paths the root/LocalSystem account uses; Windows path matches what GetStateDir() produces for LocalSystem, Linux/macOS path is correctly derived via user.Lookup("root"). |
| README.md | Documents --user flag with accurate scope (Linux/macOS) and updated privilege notes; no issues found. |
| docs/USAGE.md | Updated service management docs to describe --user flag and privilege requirements; accurate and consistent with code. |

</details>

<details open><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `cmd/utils.go`, line 52-74 ([link](https://github.com/surgedm/surge/blob/06ec69906de57db971d26420f51ee343750c0087/cmd/utils.go#L52-L74)) 

   <a href="#">P1</a> **Stale user token causes auth failure against system service**

   `root_http_server.go` removes the port file on shutdown but never removes the token file at `GetStateDir()/token`. If a user previously ran a user-level service (writing a token to `~/.local/state/surge/token`) and then switches to a system service, `readActivePort()` correctly falls through to the system port (user port file is gone), but `resolveLocalToken()` still returns the stale user token because the user token file is still present. The client then attempts to connect to the system service with the wrong token, producing a confusing authentication error. The two fallback chains need to be consistent: either check the same file for both port and token, or verify that both came from the same service instance.

   `````markdown
   Path: cmd/utils.go
   Line: 52-74

   Comment:
   **Stale user token causes auth failure against system service**

   `root_http_server.go` removes the port file on shutdown but never removes the token file at `GetStateDir()/token`. If a user previously ran a user-level service (writing a token to `~/.local/state/surge/token`) and then switches to a system service, `readActivePort()` correctly falls through to the system port (user port file is gone), but `resolveLocalToken()` still returns the stale user token because the user token file is still present. The client then attempts to connect to the system service with the wrong token, producing a confusing authentication error. The two fallback chains need to be consistent: either check the same file for both port and token, or verify that both came from the same service instance.

   `````
   </details>

2. `cmd/utils.go`, line 107-130 ([link](https://github.com/surgedm/surge/blob/fe4835f29977e343891754e91f4d7b57cbf318ed/cmd/utils.go#L107-L130)) 

   <a href="#">P1</a> **TOCTOU mismatch between port and token resolution**

   `resolveAPIConnection` calls `readActivePort()` (which internally calls `getActiveConnectionDetails()`) and then `resolveLocalToken()` (which calls `getActiveConnectionDetails()` a second time). Between the two calls, a service transition — for example the user-level service stopping and its port file being removed — can cause the port to come from the user service while the token comes from the system service, or vice versa. The resulting mismatch sends the wrong bearer token to the server, producing an opaque authentication error with no hint about which service pair was mismatched.

   The straightforward fix is to call `getActiveConnectionDetails()` once in `resolveAPIConnection`, pass the returned `tokenFile` down into token resolution, and eliminate the duplicate file reads.

   `````markdown
   Path: cmd/utils.go
   Line: 107-130

   Comment:
   **TOCTOU mismatch between port and token resolution**

   `resolveAPIConnection` calls `readActivePort()` (which internally calls `getActiveConnectionDetails()`) and then `resolveLocalToken()` (which calls `getActiveConnectionDetails()` a second time). Between the two calls, a service transition — for example the user-level service stopping and its port file being removed — can cause the port to come from the user service while the token comes from the system service, or vice versa. The resulting mismatch sends the wrong bearer token to the server, producing an opaque authentication error with no hint about which service pair was mismatched.

   The straightforward fix is to call `getActiveConnectionDetails()` once in `resolveAPIConnection`, pass the returned `tokenFile` down into token resolution, and eliminate the duplicate file reads.

   `````
   </details>

</details>

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
cmd/utils.go:107-130
**TOCTOU mismatch between port and token resolution**

`resolveAPIConnection` calls `readActivePort()` (which internally calls `getActiveConnectionDetails()`) and then `resolveLocalToken()` (which calls `getActiveConnectionDetails()` a second time). Between the two calls, a service transition — for example the user-level service stopping and its port file being removed — can cause the port to come from the user service while the token comes from the system service, or vice versa. The resulting mismatch sends the wrong bearer token to the server, producing an opaque authentication error with no hint about which service pair was mismatched.

The straightforward fix is to call `getActiveConnectionDetails()` once in `resolveAPIConnection`, pass the returned `tokenFile` down into token resolution, and eliminate the duplicate file reads.

### Issue 2 of 2
cmd/service_kardianos.go:172
The `--user` flag has no platform guard. On Windows, `kardianos/service` does not recognise the `UserService` option — it is silently ignored and the service is installed as a system service, which requires an elevated terminal. A user who runs `surge service install --user` on Windows expecting to avoid elevation will instead get an access-denied error with no indication that `--user` is a POSIX-only feature. Adding a guard that returns a clear error on Windows prevents the silent misbehaviour.

```suggestion
	serviceCmd.PersistentFlags().BoolVar(&userLevelService, "user", false, "Manage the service at the user level (no root/admin required on POSIX)")
	serviceCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
		if userLevelService && runtime.GOOS == "windows" {
			return fmt.Errorf("--user is not supported on Windows; omit the flag to manage the system service")
		}
		return nil
	}
```


`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(review): address Greptile review com..."](https://github.com/surgedm/surge/commit/fe4835f29977e343891754e91f4d7b57cbf318ed) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31479390)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,17 +115,20 @@ Surge provides an official way to manage it as a system service (daemon). This i
 # Install Surge as a system service
 surge service install
 
-# Manage the service
-surge service start
-surge service stop
-surge service status
+# Install Surge as a user-level service (Linux/macOS)
+surge service install --user
+
+# Manage the service (append --user if installed as a user service)
+surge service start [--user]
+surge service stop [--user]
+surge service status [--user]
 
 # Uninstall the service
-surge service uninstall
+surge service uninstall [--user]
 ```
 
 > [!NOTE]
-> On Linux, these commands may require `sudo`. On Windows, they should be run in an elevated (Administrator) terminal.
+> When installing as a global system service, these commands may require `sudo` on Linux/macOS, or an elevated (Administrator) terminal on Windows. Using `--user` does not require elevated privileges.
 
 ### 4. Remote TUI
 
--- a/cmd/service_kardianos.go
+++ b/cmd/service_kardianos.go
@@ -11,6 +11,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var userLevelService bool
+
 var serviceConfig = &service.Config{
 	Name:        "surge",
 	DisplayName: "Surge Download Manager",
@@ -73,7 +75,11 @@ func (p *program) Stop(s service.Service) error {
 
 func GetService() (service.Service, error) {
 	prg := &program{}
-	return service.New(prg, serviceConfig)
+	cfg := *serviceConfig
+	if userLevelService {
+		cfg.Option = service.KeyValue{"UserService": true}
+	}
+	return service.New(prg, &cfg)
 }
 
 func runAction(action func(service.Service) error, successMsg string) func(*cobra.Command, []string) error {
@@ -160,4 +166,5 @@ func init() {
 	serviceCmd.AddCommand(serviceStopCmd)
 	serviceCmd.AddCommand(serviceStatusCmd)
 	serviceCmd.AddCommand(serviceRunCmd)
+	serviceCmd.PersistentFlags().BoolVar(&userLevelService, "user", false, "Manage the service at the user level (no root/admin required on POSIX)")
 }
--- a/cmd/service_kardianos.go
+++ b/cmd/service_kardianos.go
@@ -77,7 +77,10 @@ func GetService() (service.Service, error) {
 	prg := &program{}
 	cfg := *serviceConfig
 	if userLevelService {
-		cfg.Option = service.KeyValue{"UserService": true}
+		if cfg.Option == nil {
+			cfg.Option = service.KeyValue{}
+		}
+		cfg.Option["UserService"] = true
 	}
 	return service.New(prg, &cfg)
 }
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -22,7 +22,11 @@ func readActivePort() int {
 	portFile := filepath.Join(config.GetRuntimeDir(), "port")
 	data, err := os.ReadFile(portFile)
 	if err != nil {
-		return 0
+		systemPortFile := filepath.Join(config.GetSystemRuntimeDir(), "port")
+		data, err = os.ReadFile(systemPortFile)
+		if err != nil {
+			return 0
+		}
 	}
 	var port int
 	_, _ = fmt.Sscanf(string(data), "%d", &port)
@@ -52,6 +56,21 @@ func resolveLocalToken() string {
 	if token := strings.TrimSpace(os.Getenv("SURGE_TOKEN")); token != "" {
 		return token
 	}
+	
+	stateTokenFile := filepath.Join(config.GetStateDir(), "token")
+	if data, err := os.ReadFile(stateTokenFile); err == nil {
+		if t := strings.TrimSpace(string(data)); t != "" {
+			return t
+		}
+	}
+
+	systemTokenFile := filepath.Join(config.GetSystemStateDir(), "token")
+	if data, err := os.ReadFile(systemTokenFile); err == nil {
+		if t := strings.TrimSpace(string(data)); t != "" {
+			return t
+		}
+	}
+
 	return ensureAuthToken()
 }
 
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -56,7 +56,7 @@ func resolveLocalToken() string {
 	if token := strings.TrimSpace(os.Getenv("SURGE_TOKEN")); token != "" {
 		return token
 	}
-	
+
 	stateTokenFile := filepath.Join(config.GetStateDir(), "token")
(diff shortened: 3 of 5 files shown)


SuperCoolPencil:
@greptileai this PR needs to fix the issue on both windows and linux

SuperCoolPencil:
Closing as superseded by #437, which keeps the service fallback fix paired with the extension discovery/UI fixes and passes the current pre-push checks.

PR #450, opened 2026-05-15
Fixed various issues in internal/engine/types

junaid2005p:
<h3>Greptile Summary</h3>

This PR consolidates the dual `RuntimeConfig` structs (one in `internal/config`, one in `internal/engine/types`) into a single `types.RuntimeConfig`, renames `MaxConnectionsPerHost` → `MaxConnectionsPerDownload` throughout the codebase, and introduces `DefaultRuntimeConfig()` so callers no longer create sparse struct literals that silently disable health-check features.

- **Field rename & constant adjustment**: `MaxConnectionsPerHost` → `MaxConnectionsPerDownload` across 29 files; getter fallback constant changed from `PerHostMax = 64` to `PerDownloadMax = 32`; a deprecated `MaxConnectionsPerHost` alias (tagged `json:\"-\"`) is retained in `NetworkSettings` for backward compatibility.
- **Getter semantics**: Health-check fields (`StallTimeout`, `SlowWorkerThreshold`, `SlowWorkerGracePeriod`, `SpeedEmaAlpha`) now treat zero as \"opt-out\" rather than falling back to the package default, backed by matching guards in `health.go` and `worker.go`. `Download()` and both downloader constructors now use `DefaultRuntimeConfig()` to fix the stall/slow-worker regression.
- **Bitmap normalisation**: `RestoreBitmap` now allocates a correctly-sized bitmap and copies into it, preventing a panic when a persisted bitmap is shorter than expected.

Safe to merge; all production paths now use DefaultRuntimeConfig() so health checks remain active, and bitmap normalisation prevents resume-time panics.

The core fixes are correct and well-tested. Two undocumented constant changes (SlowWorkerThreshold 0.50→0.30 and StallTimeout 5s→3s) alter engine behaviour for all default-settings users, and a minor getter asymmetry means MaxTaskRetries=0 cannot disable retries while its sibling fields can be zeroed to opt out.

internal/engine/types/config.go — the constant and getter-semantics changes are the most impactful and least documented part of this PR.

| Filename | Overview |
|----------|----------|
| internal/engine/types/config.go | Core runtime config refactor: renamed MaxConnectionsPerHost→MaxConnectionsPerDownload, PerHostMax 64→PerDownloadMax 32, changed getter semantics to allow 0 as an opt-out for health-check fields, added DefaultRuntimeConfig(), consolidated all constants into one block. SlowWorkerThreshold and StallTimeout defaults changed without documentation. |
| internal/config/settings.go | Renamed MaxConnectionsPerHost→MaxConnectionsPerDownload, kept deprecated alias with json:"-", eliminated local RuntimeConfig struct in favour of types.RuntimeConfig, updated ToRuntimeConfig(). |
| internal/engine/types/progress.go | Added bitmapLayout() helper, RestoreBitmap now normalises short/malformed bitmaps, added out-of-bounds guards in setChunkState/getChunkState. |
| internal/engine/concurrent/health.go | Added stallTimeout > 0 and threshold > 0 guards so zero values correctly disable stall detection and slow-worker cancellation. |
| internal/engine/types/config_convert.go | Deleted: no longer needed now that settings.ToRuntimeConfig() returns types.RuntimeConfig directly. |
| internal/download/manager.go | Download() now calls types.DefaultRuntimeConfig() instead of a hand-rolled sparse struct, fixing the previous stall/slow-worker regression. |

</details>

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
internal/engine/types/config.go:41-44
**Silent behavioral changes to default health-check constants**

`SlowWorkerThreshold` dropped from `0.50` to `0.30`, and `StallTimeout` dropped from `5s` to `3s`. For users relying on engine defaults (i.e. not configuring these in settings), the slow-worker cancellation threshold is now 30 % of mean speed instead of 50 %, making it considerably less aggressive. Stall detection also fires sooner (3 s vs 5 s). These are meaningful behavioral differences that aren't captured in the PR description, making them easy to overlook in production if users report changed download behaviour.

Also note: the renamed constant `PerDownloadMax = 32` (was `PerHostMax = 64`) halves the getter fallback for connections. All known production paths now call `DefaultRuntimeConfig()` which explicitly sets 32, so this only matters for partial struct literals—but the change is worth documenting.

### Issue 2 of 2
internal/engine/types/config.go:120-125
**`GetMaxTaskRetries` asymmetry with opt-out siblings**

The struct comment documents two groups: capacity-style settings (zero → use default) and opt-out settings (zero → disable). `GetMaxTaskRetries` uses `<= 0` as the fallback, which is correct for a capacity-style field, but that means `MaxTaskRetries = 0` cannot be used to disable retries entirely. `StallTimeout`, `SlowWorkerThreshold`, `SlowWorkerGracePeriod`, and `SpeedEmaAlpha` all allow 0 to opt out. If a future caller wants zero retries (e.g. a probe-only path), they have no way to express that. A comment on the getter clarifying this constraint would close the ambiguity.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix: Fixed sparse runtime configs"](https://github.com/surgedm/surge/commit/7c4282b51d0056be6da04c11b6b08a7bcff14977) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32266256)</sub>


SuperCoolPencil (approved):
LGTM.

PR #458, opened 2026-05-16
test: refactor shared state DB reset into TestMain for cmd tests

terminalchai:
## Problem

The \cmd\ package tests had inconsistent state DB teardown across test helpers. \setupXDGEnvIsolation\ in \	est_env_test.go\ called \state.CloseDB()\ directly in its cleanup function, but \TestMain\ only set \XDG_CONFIG_HOME\ and \APPDATA\/\USERPROFILE\ — leaving \XDG_DATA_HOME\, \XDG_STATE_HOME\, \XDG_CACHE_HOME\, \XDG_RUNTIME_DIR\, and \HOME\ pointing at real user directories. This meant:

- The shared suite DB was initialised against real XDG paths, not the isolated temp dir.
- Tests that restored XDG isolation via \state.CloseDB()\ could leave the suite-level DB in a closed state without reopening it against the temp dir, causing subsequent tests in the same run to fail non-deterministically.

## Fix

1. **\TestMain\**: set all six XDG env vars (+ \HOME\) to the temp dir, then call a new \
esetSharedStateDB()\ helper that closes any existing DB, ensures config dirs exist, and reopens the DB against the isolated temp path. Also close the DB before \os.RemoveAll\ to avoid Windows file-lock errors.

2. **\
esetSharedStateDB()\**: single source of truth for suite-level DB reset — used by both \TestMain\ and the per-test cleanup in \setupXDGEnvIsolation\.

3. **\setupXDGEnvIsolation\**: replace the bare \state.CloseDB()\ call with \
esetSharedStateDB()\ so the suite DB is always left in a known-open state after each per-test isolation window closes.

## Testing

\\\
go test ./cmd/... -count=1 -race
\\\

All tests pass.

<h3>Greptile Summary</h3>

This PR applies two correctness fixes to the test setup helpers in `cmd/main_test.go`. Both changes tighten error-handling paths that could previously leave the test suite in a broken or partially-initialised state.

- `resetSharedStateDB` now returns early when `config.EnsureDirs()` fails instead of proceeding to call `state.Configure` against a directory that was never created.
- `TestMain` now removes the allocated temp directory before calling `os.Exit(1)` on a `resetSharedStateDB` failure, preventing stale directories from being left behind.

Safe to merge — the changes are confined to test infrastructure, fix two well-scoped error-handling gaps, and do not touch production code paths.

Both changes are isolated to test setup helpers. The resetSharedStateDB guard prevents a spurious state.Configure call when directory creation fails, and the os.RemoveAll addition ensures the temp directory is cleaned up on every exit path. No logic regressions are introduced.

| Filename | Overview |
|----------|----------|
| cmd/main_test.go | Two targeted fixes: early-return guard in `resetSharedStateDB` so `state.Configure` is not called after a failed `EnsureDirs`, and cleanup of `tmpDir` on the `resetSharedStateDB` error exit path in `TestMain`. |

</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/cmd-state-d..."](https://github.com/surgedm/surge/commit/de12e0e449d43e8a4cde2493860e0a577a6593e2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32465079)</sub>


SuperCoolPencil:
I don't understand how this PR has 0 changes but 5 commits. Am I going crazy? lol

SuperCoolPencil:
> I don't understand how this PR has 0 changes but 5 commits. Am I going crazy? lol

Yep my bad. Made a wrong merge commit

SuperCoolPencil (approved):

SuperCoolPencil:
Thanks for contributing :)

PR #471, opened 2026-06-02
fix: batch downloads were being overwritten and only last one would be queued. Fixes #469

### LINKED ISSUE #469: Batch importing download links doesn't work because of the confirmation dialog
**Describe the bug**
When I try to batch import links into the queue it fails to import them because of the confirmation dialog whether I do it from the TUI or from CLI. And I couldn't find anything related to confirmation dialog in settings.

**To Reproduce**

1. Open surge TUI
2. From another terminal run `surge add -b linksfile`
3. Go back to the surge TUI and you'll see only the latest link's confirmation dialog showing.

**Expected behavior**

- Don't show confirmation dialog at all
- Show one confirmation dialog for all of the links maybe?

**Screenshots**

- OS: Arch Linux
- Surge Version: Surge v0.8.6-1
- Installed From: aur/surge-bin

**Additional context**

junaid2005p:
<h3>Greptile Summary</h3>

This PR fixes the core bug where concurrent `DownloadRequestMsg` events overwrote each other in the TUI, leaving only the last one in the confirmation queue. It also introduces a `/download/batch` HTTP endpoint and a `--confirm` CLI flag for grouped batch approval.

- **Core fix**: `handleDownloadRequestMsg` and `handleBatchDownloadRequestMsg` now enqueue incoming requests into `pendingRequestQueue` / `pendingBatchRequestQueue` when the TUI is already in a confirmation state; all exit paths (`cancel`, `confirm`, `focus`) call `showNextPendingRequest()` to drain the queue.
- **New batch endpoint**: `POST /download/batch` accepts a `BatchDownloadRequest` with either TUI-confirmation (`skip_approval=false`) or direct enqueue (`skip_approval=true`), with partial-failure reporting via `207 Multi-Status`.
- **New `--confirm` flag**: `surge add --batch file.txt --confirm` routes the batch through the new endpoint for a single grouped TUI confirmation instead of N individual prompts.

The fix is well-scoped and the new queuing logic is covered by targeted tests; the findings are non-blocking suggestions that do not affect the main fix.

The core overwrite bug is correctly addressed through a clean queue/drain pattern applied consistently to all confirmation exit paths. The new batch endpoint mirrors the existing single-download handler structure, including partial-failure handling and activeDownloads accounting. Tests cover the queuing state machine for both single and batch requests. The remaining gaps are low-consequence edge cases that do not affect the primary fix.

internal/tui/download_requests.go (queuing guard misses file-picker states) and cmd/utils.go (sendBatchToServer response handling)

| Filename | Overview |
|----------|----------|
| internal/tui/download_requests.go | New file introducing queued dispatch for single and batch download requests; the queuing guard omits FilePickerState and BatchFilePickerState |
| cmd/root_downloads.go | Adds handleBatchDownload with per-item validation, partial-failure reporting, and correct activeDownloads accounting mirroring handleDownload |
| cmd/utils.go | Adds sendBatchToServer and sendToServerWithApproval; sendBatchToServer does not accept 207 Multi-Status as a success response |
| internal/tui/update_modals.go | Batch confirm now processes both pendingBatchRequests and pendingBatchURLs, calls showNextPendingRequest on all exit paths |
| internal/tui/update_filepicker.go | Sets inputs[2] to default download path on file selection; does not clear pendingBatchRequests before entering BatchConfirmState |
| cmd/http_api_test.go | Adds publishRecordingHTTPService, batchAddRecordingService, and two tests covering confirmation and partial-failure batch paths |
| cmd/add.go | Adds --confirm flag; batch+confirm goes through new batch endpoint, individual downloads use sendToServerWithApproval with skipApproval=!confirm |
| internal/engine/events/events.go | Adds BatchDownloadRequestMsg struct, EventTypeBatchRequest constant, and SSE encode/decode support |
| internal/tui/model.go | Adds pendingBatchRequests, pendingRequestQueue, and pendingBatchRequestQueue fields to RootModel |
| internal/tui/update_test.go | Adds queuing tests for both single and batch download request messages; good state-machine coverage |
| internal/tui/view.go | BatchConfirmState view now distinguishes source file from output path; shows shared path when batchFilePath matches inputs[2] |
| internal/tui/keys.go | Batch confirm keys changed from y/n/enter/esc to enter/esc only |
| internal/tui/update_events.go | DownloadRequestMsg and BatchDownloadRequestMsg handling delegated to new handler functions; logic moved cleanly |
| cmd/http_api.go | Registers /download/batch route consistently with existing /download route pattern |
| internal/tui/update_input.go | Extension confirmation now calls showNextPendingRequest on both confirm and cancel paths |

</details>

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
internal/tui/download_requests.go:61-62
The queuing guard doesn't include `FilePickerState` or `BatchFilePickerState`. If a `BatchDownloadRequestMsg` arrives while the user is actively navigating either file picker, it immediately transitions state to `BatchConfirmState`, abandoning the in-progress selection. For `DownloadRequestMsg` a missed guard during the file picker is relatively benign (the download just starts silently), but a batch message is always modal — it always replaces the visible state — making the interruption visible and disruptive.

```suggestion
func (m RootModel) handleBatchDownloadRequestMsg(msg events.BatchDownloadRequestMsg, queueIfBusy bool) (tea.Model, tea.Cmd) {
	if queueIfBusy && (m.state == ExtensionConfirmationState || m.state == DuplicateWarningState || m.state == BatchConfirmState || m.state == FilePickerState || m.state == BatchFilePickerState) {
```

### Issue 2 of 3
internal/tui/update_filepicker.go:17-19
`handleBatchFileSelection` doesn't clear `m.pendingBatchRequests` before entering `BatchConfirmState`. If a `BatchDownloadRequestMsg` arrived just before the user confirmed a file selection (possible while `FilePickerState` / `BatchFilePickerState` are not yet in the queuing guard), both `pendingBatchURLs` and `pendingBatchRequests` would be non-nil simultaneously, causing `updateBatchConfirm` to process both batches and `view.go` to report the wrong URL count.

```suggestion
	m.pendingBatchURLs = urls
	m.pendingBatchRequests = nil
	m.inputs[2].SetValue(m.defaultDownloadPath())
	m.batchFilePath = path
```

### Issue 3 of 3
cmd/utils.go:256-264
`sendBatchToServer` only accepts `200 OK` or `202 Accepted` as success. When `skipApproval=true` and some items fail, the server returns `207 Multi-Status`, which this code treats as an error — even though downloads were partially queued. The function already accepts a `skipApproval bool` parameter, so any caller that passes `true` would incorrectly surface a failure despite real work being done.

```suggestion
	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusMultiStatus {
		body, _ := io.ReadAll(resp.Body)
		return fmt.Errorf("server error: %s - %s", resp.Status, string(body))
	}

	return nil
}

// GetRemoteDownloads fetches all downloads from the running server
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix: addressed greptile concerns"](https://github.com/surgedm/surge/commit/a72e69d03ad0fe2123da0d54460ed9093d8660b1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35053936)</sub>

### DIFF
--- a/internal/tui/keys.go
+++ b/internal/tui/keys.go
@@ -418,12 +418,12 @@ var Keys = KeyMap{
 	},
 	BatchConfirm: BatchConfirmKeyMap{
 		Confirm: key.NewBinding(
-			key.WithKeys("y", "Y", "enter"),
-			key.WithHelp("y", "confirm"),
+			key.WithKeys("enter"),
+			key.WithHelp("enter", "confirm"),
 		),
 		Cancel: key.NewBinding(
-			key.WithKeys("n", "N", "esc"),
-			key.WithHelp("n", "cancel"),
+			key.WithKeys("esc"),
+			key.WithHelp("esc", "cancel"),
 		),
 	},
 	Update: UpdateKeyMap{
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -20,6 +20,7 @@ var addCmd = &cobra.Command{
 
 		batchFile, _ := cmd.Flags().GetString("batch")
 		output, _ := cmd.Flags().GetString("output")
+		confirm, _ := cmd.Flags().GetBool("confirm")
 
 		var urls []string
 		urls = append(urls, args...)
@@ -44,6 +45,14 @@ var addCmd = &cobra.Command{
 		}
 		resolvedOutput := resolveClientOutputPath(output)
 
+		if batchFile != "" && confirm {
+			if err := sendBatchToServer(urls, resolvedOutput, baseURL, token, false); err != nil {
+				return err
+			}
+			fmt.Printf("Batch confirmation requested for %d downloads.\n", len(urls))
+			return nil
+		}
+
 		// Send downloads to server
 		count := 0
 		attempted := 0
@@ -53,7 +62,7 @@ var addCmd = &cobra.Command{
 				continue
 			}
 			attempted++
-			if err := sendToServer(url, mirrors, resolvedOutput, baseURL, token); err != nil {
+			if err := sendToServerWithApproval(url, mirrors, resolvedOutput, baseURL, token, !confirm); err != nil {
 				fmt.Printf("Error adding %s: %v\n", url, err)
 				continue
 			}
@@ -77,4 +86,5 @@ func init() {
 	rootCmd.AddCommand(addCmd)
 	addCmd.Flags().StringP("batch", "b", "", "File containing URLs to download (one per line)")
 	addCmd.Flags().StringP("output", "o", "", "Output directory (defaults to current working directory)")
+	addCmd.Flags().Bool("confirm", false, "Show confirmation prompt before starting downloads")
 }
--- a/cmd/http_api.go
+++ b/cmd/http_api.go
@@ -36,6 +36,10 @@ func registerHTTPRoutes(mux *http.ServeMux, port int, defaultOutputDir string, s
 		handleDownload(w, r, defaultOutputDir, service)
 	})
 
+	mux.HandleFunc("/download/batch", func(w http.ResponseWriter, r *http.Request) {
+		handleBatchDownload(w, r, defaultOutputDir, service)
+	})
+
 	mux.HandleFunc("/pause", requireMethod(http.MethodPost, withRequiredID(func(w http.ResponseWriter, _ *http.Request, id string) {
 		if err := service.Pause(id); err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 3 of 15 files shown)


internal/tui/keys.go
SuperCoolPencil, line 1: looks like a regression?

SuperCoolPencil (approved):

PR #481, opened 2026-06-10
fix(cli): don't split a single URL into mirrors on query-string commas (#316)

### LINKED ISSUE #316: Surge doesn't parse archive.org compress links correctly.
**Describe the bug**

When I want to download a complete entry from archive.org I use the "x Original" files link. This works with aria2c for example, but not with Surge.

**To Reproduce**

Go to an archive.org entry. For example: https://archive.org/details/live-image-bookworm-mate-20260403-amd64.hybrid

In the "Show all" section copy the "10 Original" download link.

Paste it into Surge or append it to Surge in a command line (eg: "surge https://archive.org/compress/live-image-bookworm-mate-20260403-amd64.hybrid/formats=PNG,ITEM%20TILE,LOG,ISO%20IMAGE,ARCHIVE%20BITTORRENT,METADATA,UNKNOWN,TEXT"

**Expected behavior**
Surge should parse the url correctly and download a zip file containing all the contents. Here it downloaded a 72kb zip file containing a single png screenshot. Using the same link with aria2c gets a zip file with all the files.

**Logs**
Surge automatically writes debug log files.

~/.local/state/surge/logs/ is empty here.

- OS: debian sid
- Surge Version: Surge v0.7.6
- Installed From: GitHub Release

StressTestor:
Fixes #316.

`surge` mangles any URL that has commas in its query string. archive.org "compress" links are the case in the issue:

```
surge "https://archive.org/compress/<item>/formats=PNG,ITEM TILE,LOG,ISO IMAGE"
```

`ParseURLArg` splits the argument on every comma to support comma-separated mirrors, so that one URL gets cut into `.../formats=PNG` plus three junk "mirrors" (`ITEM TILE`, `LOG`, `ISO IMAGE`). only the first fragment downloads - a tiny stub instead of the archive. aria2c handles the same link fine.

## the fix

a comma only starts a new mirror when the next segment parses as an absolute http(s) URL. otherwise the comma lived inside the previous URL, so it gets glued back. surge only downloads http/https (`internal/clipboard/validator.go`, `cmd/connect.go`), so a schemeless segment is never a standalone mirror.

genuine mirror lists are unaffected, since every real mirror is a full URL:

```
surge "http://a.com/file,http://b.com/file,http://c.com/file"   # still 3 mirrors
```

## tests

extended `TestParseURLArg_Unit` in `cmd/mirrors_integration_test.go`:

- the archive.org link stays one URL
- a query-comma URL followed by a real mirror splits into exactly those two

red before the change (the archive case shredded into 4 mirrors), green after. the existing cases and `TestMirrors_CLI_Integration` are unchanged. `go test ./...`, `go build ./...`, `go vet`, and golangci-lint v2 are all clean.

<h3>Greptile Summary</h3>

This PR fixes query-string commas being misinterpreted as mirror separators in `ParseURLArg`. It replaces the naive split-on-every-comma approach with a `startsNewMirror` helper that only opens a new mirror slot when a segment starts with the literal `http://` or `https://` prefix.

- `cmd/utils.go`: Introduces `startsNewMirror` and updates `ParseURLArg` to glue non-http segments back to the preceding URL rather than treating them as standalone mirrors.
- `cmd/mirrors_integration_test.go`: Adds three new table-driven test cases — the archive.org formats link, a mixed query-comma-plus-real-mirror input, and the bare-scheme (`http:`) edge case — all of which were red before this fix.

This is a small, self-contained fix to a parsing function with no side effects on download logic; genuine multi-mirror inputs are unaffected, and the new tests confirm both the happy path and the tricky edge cases.

The change is minimal and surgical: a two-line helper and a guard in one loop. The fix uses a `strings.HasPrefix` check rather than `url.Parse`, correctly handling the bare-scheme edge case raised in the previous review. All pre-existing test cases remain green, and the three new cases directly cover the bug scenario from the issue report as well as the boundary cases.

| Filename | Overview |
|----------|----------|
| cmd/utils.go | Adds `startsNewMirror` helper and reworks the comma-split loop in `ParseURLArg` so that only segments starting with `http://` or `https://` open a new mirror slot; all other segments are glued back to the preceding URL. |
| cmd/mirrors_integration_test.go | Extends `TestParseURLArg_Unit` with three new table cases: query-string-only URL, query-comma URL mixed with a real mirror, and bare-scheme segment gluing — all covering the edge cases raised in the original issue and the previous review thread. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(cli): match the http(s):// prefix in..."](https://github.com/surgedm/surge/commit/d1bb68aa311bff0ce683837d765dec947dc7b061) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36750195)</sub>

### DIFF
--- a/cmd/mirrors_integration_test.go
+++ b/cmd/mirrors_integration_test.go
@@ -122,6 +122,12 @@ func TestParseURLArg_Unit(t *testing.T) {
 			expectedMirrors: []string{"https://archive.org/compress/item/formats=PNG,LOG", "http://mirror.example.com/item.zip"},
 		},
 		{
+			name:            "Bare scheme in query is not a mirror boundary",
+			input:           "https://primary.com/file?a=1,http:,http://mirror.example.com/file",
+			expectedURL:     "https://primary.com/file?a=1,http:",
+			expectedMirrors: []string{"https://primary.com/file?a=1,http:", "http://mirror.example.com/file"},
+		},
+		{
 			name:            "Empty URL",
 			input:           "",
 			expectedURL:     "",
--- a/cmd/mirrors_integration_test.go
+++ b/cmd/mirrors_integration_test.go
@@ -110,6 +110,18 @@ func TestParseURLArg_Unit(t *testing.T) {
 			expectedMirrors: []string{"http://a.com", "http://b.com"},
 		},
 		{
+			name:            "Single URL with commas in query string (archive.org formats)",
+			input:           "https://archive.org/compress/item/formats=PNG,ITEM TILE,LOG,ISO IMAGE",
+			expectedURL:     "https://archive.org/compress/item/formats=PNG,ITEM TILE,LOG,ISO IMAGE",
+			expectedMirrors: []string{"https://archive.org/compress/item/formats=PNG,ITEM TILE,LOG,ISO IMAGE"},
+		},
+		{
+			name:            "Query-comma URL followed by a real mirror",
+			input:           "https://archive.org/compress/item/formats=PNG,LOG,http://mirror.example.com/item.zip",
+			expectedURL:     "https://archive.org/compress/item/formats=PNG,LOG",
+			expectedMirrors: []string{"https://archive.org/compress/item/formats=PNG,LOG", "http://mirror.example.com/item.zip"},
+		},
+		{
 			name:            "Empty URL",
 			input:           "",
 			expectedURL:     "",
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -86,14 +86,33 @@ func ParseURLArg(arg string) (string, []string) {
 	parts := strings.Split(arg, ",")
 	var urls []string
 	for _, p := range parts {
-		if trimmed := strings.TrimSpace(p); trimmed != "" {
-			urls = append(urls, trimmed)
+		trimmed := strings.TrimSpace(p)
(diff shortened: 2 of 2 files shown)


StressTestor:
good catch. `url.Parse` treats a bare `http:` / `https:` as a valid scheme, so a query value like `?schemes=http:,https:` would get split off as a bogus mirror. switched to a literal `http://` / `https://` prefix check, which also matches how `internal/clipboard/validator.go` recognizes URLs, and added a regression test for the bare-scheme case. pushed in d1bb68a.

junaid2005p:
LGTM 
Thank you so much for contributing @StressTestor 

Also thanks for using surge!!

junaid2005p (approved):

StressTestor:
appreciate it man. glad it was a clean one. surge is a solid tool, i'll probably be back with more

SuperCoolPencil:
Oh, this solves #316 
Thank you so much @StressTestor 
Wasn't able to solve this one before.

StressTestor:
appreciate it man. the sneaky part was the comma doing double duty - it's the mirror separator and it also shows up inside archive.org's `formats=` query, so a plain split couldn't tell a real second url from a comma inside one. the http(s):// prefix check is what disambiguates. fun one to track down

PR #502, opened 2026-06-17
replace sqlite with gob

SuperCoolPencil:
- **refactor: replace go-humanize dependency with custom implementation**
- **refactor(state): replace sqlite backend with custom gob storage**
- **feat(state): restore NormalizeStaleDownloads and ValidateIntegrity using gob backend**
- **fix(state): handle gob nil slices and Task atomic fields, repair test suite**
- **fix: resolve failing tests and fix missing master list updates in gob backend**
- **refactor: remove obsolete database unit tests**
- **docs: remove remaining obsolete sqlite references from code comments**
- **test: seed master list with download entries in integration tests**
- **test: fix broken test fixtures missing master list seeding and handle decimal changes in size formatter**

<h3>Greptile Summary</h3>

This PR replaces the SQLite-backed persistence layer (`modernc.org/sqlite`) with a custom gob-based file store, and replaces the `go-humanize` direct dependency with a small custom formatter. All changed files have been updated to match the new two-file layout: a `master.gob` index and per-download `details/<id>.gob` files, with atomic writes via `os.CreateTemp` + `os.Rename`.

- **State backend rewrite**: `db.go` introduces `atomicWrite`, `loadGob`, `ensureDirs`, and `cleanupOrphans`; `state.go` is fully rewritten around `MasterState`/`DetailState` gob structs with proper `sync.RWMutex` locking.
- **Bug fixes from prior review**: silent master-list error discards, `UpdateDefaultRateLimit` override bug, `GobDecode` silent nil-return, `LoadStates` error swallowing, `cleanupOnce` data race, and stale `sql.ErrNoRows` in tests are all addressed.
- **Test suite updated**: All integration and unit tests now seed the master list explicitly via `AddToMasterList` before calling `SaveState`, matching the new two-step API contract.

The refactor is largely complete and the most critical bugs flagged in the previous review round have been addressed; the remaining concern is the implicit two-step caller contract (AddToMasterList must precede SaveState) which is easy to violate silently in future call sites.

All previously-flagged blocking issues — silent master-list overwrites, error discard in DeleteTasks, GobDecode nil on corrupt data, LoadStates dropping errors, cleanupOnce data race, and the stale sql.ErrNoRows test assertion — are fixed in this diff. The gob write path uses atomic rename, locking is consistent, and the test suite has been updated to seed the master list correctly. The main residual risk is the undocumented ordering requirement between AddToMasterList and SaveState: SaveState silently no-ops the master-list update (logging only at debug level) when AddToMasterList has not been called first, and LoadState will then return os.ErrNotExist even though the detail gob was written. This contract is not enforced or surfaced as an error to the caller, making it easy for future callers to produce orphaned detail files.

internal/engine/state/state.go (SaveState/AddToMasterList ordering contract) and internal/engine/state/db.go (ensureDirs/cleanupOrphans interaction) deserve the most careful review.

| Filename | Overview |
|----------|----------|
| internal/engine/state/db.go | Core gob backend: replaces SQL connection pool with atomic file writes; previous data-race on cleanupOnce now fixed with cleanupMu; initDB / cleanupOrphans production-reachability concern now addressed via ensureDirs. |
| internal/engine/state/state.go | Main state API; previous silent-error discards, UpdateDefaultRateLimit override bug, and master-list empty-truncation issues are all fixed; SaveState still silently no-ops master-list update when AddToMasterList was not called first (by design, documented via debug log). |
| internal/engine/state/state_test.go | Stale database/sql import and sql.ErrNoRows check in TestDeleteState have been fixed; test now uses errors.Is(err, os.ErrNotExist) correctly. |
| internal/engine/types/models.go | GobEncode/GobDecode added for Task to handle atomic.Int64 field; GobDecode now correctly returns an error for corrupt data (< 16 bytes) instead of silently returning nil. |
| internal/utils/size_converter.go | Custom go-humanize replacement; index into "kMGTPE" is safe in practice — int64 max (~9.2 EB) can only reach index 5 ('E'), so the out-of-bounds concern is not reachable with the current type. |
| go.mod | Removes modernc.org/sqlite and related transitive deps; go-humanize moved from direct to indirect (something in the transitive dep tree still references it). |
| internal/core/local_service_test.go | Error from AddToMasterList is silently discarded with _ = on the seeding line; if AddToMasterList fails, the delete test may pass for the wrong reason. |

</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `internal/engine/state/state_test.go`, line 4-12 ([link](https://github.com/surgedm/surge/blob/1d12151dbe176265013ba4f8e687cb7f7cfc5b63/internal/engine/state/state_test.go#L4-L12)) 

   <a href="#">P1</a> **Stale `database/sql` import breaks `TestDeleteState`'s error assertion**

   `state_test.go` still imports `"database/sql"` (line 5) and the `TestDeleteState` switch (line 265–272) checks `case sql.ErrNoRows:` as the "acceptable" outcome. The gob backend never returns `sql.ErrNoRows` — after `DeleteState` removes the detail gob file, `LoadState` returns a wrapped `os.ErrNotExist` error (via `"failed to load detail state: %w"`). Neither `nil` nor `sql.ErrNoRows` fires, so the switch falls through silently for *any* non-nil error, including unexpected ones like a permission failure. The import should be dropped and the check replaced with `errors.Is(err, os.ErrNotExist)` or simply `if err == nil { t.Error(...) }`.

   `````markdown
   Path: internal/engine/state/state_test.go
   Line: 4-12

   Comment:
   **Stale `database/sql` import breaks `TestDeleteState`'s error assertion**

   `state_test.go` still imports `"database/sql"` (line 5) and the `TestDeleteState` switch (line 265–272) checks `case sql.ErrNoRows:` as the "acceptable" outcome. The gob backend never returns `sql.ErrNoRows` — after `DeleteState` removes the detail gob file, `LoadState` returns a wrapped `os.ErrNotExist` error (via `"failed to load detail state: %w"`). Neither `nil` nor `sql.ErrNoRows` fires, so the switch falls through silently for *any* non-nil error, including unexpected ones like a permission failure. The import should be dropped and the check replaced with `errors.Is(err, os.ErrNotExist)` or simply `if err == nil { t.Error(...) }`.

   `````
   </details>
</details>

<sub>Reviews (10): Last reviewed commit: ["refactor: improve state loading error ha..."](https://github.com/surgedm/surge/commit/7243334632379e15d156d0938859618c45d21d98) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38073786)</sub>


SuperCoolPencil:
we reduce ~3.66MB of binary size.

PR #573, opened 2026-08-01
fix/phantom progress on 403

SuperCoolPencil:
- **fix: defer activeTasks removal on error and handle permanent HTTP status codes in worker execution**
- **fix(strategy/concurrent): prevent phantom progress and handle 403 as permanent error**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved download state snapshots after failed tasks so outstanding bytes are preserved accurately.
  - Ensured permanent HTTP failures return a specific permanent-error classification.
  - Improved pause handling during retry backoff, including pause notifications and accurate zero-byte state reporting.
  - Improved task recovery and cleanup after failed or completed work.
  - Preserved shared download limits when resuming remaining work.

- **Tests**
  - Added coverage for pausing downloads while waiting to retry.
  - Added coverage for preserving shared download limits during task recovery.

### DIFF
--- a/internal/strategy/concurrent/concurrent_test.go
+++ b/internal/strategy/concurrent/concurrent_test.go
@@ -818,7 +818,6 @@ func TestConcurrentDownloader_Download_BootstrapFail_InvalidRange(t *testing.T) 
 	}
 }
 
-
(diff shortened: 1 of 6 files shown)


SuperCoolPencil:
@coderabbitai resume

Issue #579, opened 2026-08-03
Concurrent downloader fails signed URLs due to extra remainder and hedged range requests


need-tech-support:
**Describe the bug**

(This bug report is written with assistance from an AI assistant)

The concurrent downloader fails with signed/request-sensitive range URLs after the primary ranges have downloaded successfully because it makes additional HTTP requests.

In the observed download, Surge selected three workers. The three primary ranges completed, but the 4 KiB-aligned chunk calculation left a small fourth remainder task. As workers became idle, the balancer also issued speculative hedged requests for the remaining parts of active ranges. The server returned HTTP 404 for these additional requests. Surge retried the 404 and then marked the download as failed with only the small remainder task incomplete.

The initial URL was valid: the metadata probe returned `206 Partial Content`, range support and a valid total size, and all three primary ranges downloaded successfully. The attached log has been sanitized: the filename, download token, cookie, referer and local path have been replaced with non-sensitive values. The service hostname and byte counts have been retained because they are relevant to reproducing the request pattern.

**To Reproduce**

1. Use the browser extension to add an authenticated, signed or request-sensitive URL from a server that supports HTTP ranges.
2. Use concurrent downloading with three workers and the default dial hedging configuration.
3. Download a file whose size is not evenly divisible by the aligned worker chunk size.
4. Wait for the primary ranges to approach completion.
5. Observe Surge create a small remainder task and speculative hedged requests after workers become idle.
6. If the server rejects these additional requests, observe `unexpected status: 404: permanent HTTP error` and the entire download enter the error state despite the primary ranges having completed.

This can be reproduced without a private download by using a mock range server that permits the `bytes=0-0` metadata probe and N primary data ranges, but returns 404 for additional or duplicate data-range requests.

Take note this issue occurs in Rapidgator, and using aria2c for the same link yields no issues.

**Expected behavior**

With N workers, Surge should divide the initial download into exactly N ranges, adding any division/alignment remainder to the final range instead of creating another HTTP request.

A failed speculative hedge should not fail the download if the original worker successfully completes the same range. There should also be a setting to disable runtime `HedgeWork` independently of `dial_hedge_count`, which currently controls connection prewarming only.

An error classified as `ErrPermanentHTTP` should not pass through the generic worker retry loop.

**Screenshots**

Not attached because the download list displays the private filename. The debug log contains the relevant event sequence.

**Logs**

Attached: `debug.log`.

[...]

PR #550, opened 2026-07-24
fix(tls): thread --tls-ca-file and --insecure-tls through the download engine

SuperCoolPencil:
## Problem

Fixes #549.

`--tls-ca-file` and `--insecure-tls` were wired into the remote-API HTTP client (`service.NewHTTPTransport`) but not into the `NetworkPool` used by the probe and download engines. The transport was created with a bare `http.Transport` and no `TLSClientConfig`, silently ignoring both flags for every actual download.

Alpine Linux's `ca-certificates` package does not include the `Microsoft TLS RSA Root G2` root, so downloading from `download.microsoft.com` (Windows SDK, etc.) fails with:

```
tls: failed to verify certificate: x509: certificate signed by unknown authority
```

## What changed

| File | Change |
|---|---|
| `internal/utils/tls.go` *(new)* | `BuildTLSConfig(caFile, insecure)` — shared helper that **appends** to the system pool rather than replacing it |
| `internal/service/http_client.go` | `newTLSConfig` delegates to `utils.BuildTLSConfig` — removes 30 lines of duplication |
| `internal/types/config.go` | `TLSCAFile` / `TLSInsecure` added to `RuntimeConfig`; `WithTLS()` overlay helper |
| `internal/transport/network.go` | Pool key includes TLS fields; `AcquireTransport` accepts 2 new params; `createNewTransport` sets `TLSClientConfig` |
| `internal/probe/probe.go` | Forwards TLS fields from `runCfg` to `AcquireTransport` |
| `internal/strategy/concurrent/downloader.go` | `setupNetwork()` forwards TLS from `d.Runtime` |
| `internal/strategy/single/downloader.go` | `Download()` forwards TLS from `d.Runtime` |
| `internal/orchestrator/manager.go` | `DownloadRequest` carries the flags; overlaid via `WithTLS()` at probe and download-record build |
| `internal/scheduler/manager.go` | Mirror-probe inline config copies TLS fields |
| `cmd/root_downloads.go` | Both `DownloadRequest` constructions inject CLI globals |
| `docs/USAGE.md` | Global flags table + new **TLS / Certificate Troubleshooting** section |

## Key design choices

- **Append, not replace**: `BuildTLSConfig` calls `x509.SystemCertPool()` and appends the user's PEM. Adding the Microsoft root doesn't break verification for any other host.
- **Pool-keyed by TLS config**: Downloads with different CA files or insecure flags get separate transport instances — correct isolation, no cross-contamination.
[...]

PR #581, opened 2026-08-04
fix(concurrent): avoid extra initial remainder request

snowyukitty:
## Summary

- cap new parallel downloads at one initial range per actual worker
- fold division/alignment remainder into the final primary range
- leave saved resume tasks and sequential chunking unchanged

## Why

The aligned chunk size can leave a small (N+1)th remainder task after N primary ranges. Request-sensitive signed URLs can reject that extra HTTP range request even though the primary ranges are valid.

## Testing

- `go test ./internal/strategy/concurrent -run '^(TestInitialRangesStayWithinWorkerRequestBudget|TestTaskRangeAssignment)$' -count=1`
- `go test ./internal/strategy/concurrent -count=1`
- `go vet ./...`
- `go test ./internal/lint/... -count=1`
- `go build ./...`
- all Go packages except `cmd` and `internal/probe` (the two reproduced Windows baseline failures)

`go test ./... -count=1 -timeout 90s` passed every other package. On this Windows host, `cmd` still fails its dummy batch-editor test and `internal/probe` still reports a DNS resolver goroutine; both failures reproduce on the audited base. The race detector was unavailable because the host has no CGO toolchain, WSL, or running Docker daemon.

## Coordination

This patch touches `internal/strategy/concurrent/downloader.go` alongside open #550 and #580. Both current merge trees are clean, and the concurrent package passes on each combined tree.

Addresses the initial partitioning case in #579. Runtime hedging, permanent HTTP retry handling, and connection prewarming remain out of scope.

Assistance disclosure: This contribution was developed with AI assistance (OpenAI Codex).

## Summary by CodeRabbit

* **Performance**
  * Improved parallel download task allocation to respect the configured worker limit.
  * Reduced unnecessary task creation while preserving complete file coverage.
  * Improved handling of uneven file sizes, small downloads, and single-worker transfers.

* **Bug Fixes**
  * Corrected byte-range assignment for concurrent and signed-range downloads.
  * Ensured remainder data is included in the final download task without creating extra requests.
  * Ensured downloads fully cover the requested file without invalid or empty ranges.

### DIFF
--- a/internal/strategy/concurrent/chunk_test.go
+++ b/internal/strategy/concurrent/chunk_test.go
@@ -23,6 +23,7 @@ func TestTaskRangeAssignment(t *testing.T) {
 		numConns  int
 		wantChunk int64
 		wantTasks int
+		wantLast  int64
 	}{
 		{
 			name:      "Exact division",
@@ -30,6 +31,7 @@ func TestTaskRangeAssignment(t *testing.T) {
 			numConns:  4,
 			wantChunk: 25 * utils.MiB,
 			wantTasks: 4,
+			wantLast:  25 * utils.MiB,
 		},
 		{
 			name:     "Uneven division",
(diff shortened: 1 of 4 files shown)


SuperCoolPencil:
Thanks so much @snowyukitty 🚀

Issue #591, opened 2026-08-08
Bug: Errored downloads show as completed with 0% progress after restart


SuperCoolPencil:
When the Surge application is restarted, downloads that were previously in an 'errored' state are incorrectly displayed as 'completed' with 0% progress. They should instead retain and accurately display their 'errored' state upon restart.

superGekFordJ:
At first glance, this seemed like it might be related to PR #580, so I investigated it. It turns out the engine layer's backend persistence seems to be working fine (it correctly saves `Status: "error"`)

So I had my AI help investigate the TUI layer, and we found that the problem likely stems from the error state not being passed through properly. Specifically, in `internal/tui/model.go` (`InitialRootModel`), the `case "error":` block sets `dm.done = true` and `dm.started = true`, but it never populates `dm.err`. 

Here is a potential patch that might fix the issue, perhaps you could give it a try:
```patch
diff --git a/internal/tui/model.go b/internal/tui/model.go
index b4ee627..de99c4b 100644
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1,6 +1,8 @@
 package tui
 
 import (
+	"errors"
+
 	"github.com/SurgeDM/Surge/internal/orchestrator"
 	engineprogress "github.com/SurgeDM/Surge/internal/progress"
 
@@ -412,6 +414,11 @@ func InitialRootModel(serverPort int, currentVersion string, service service.Dow
 				case "error":
 					dm.done = true
 					dm.started = true
+					if s.Error != "" {
+						dm.err = errors.New(s.Error)
+					} else {
+						dm.err = errors.New("download failed")
+					}
 				case "pausing":
 					dm.pausing = true
 					dm.started = true
```

While looking into this, I also noticed a related gap: the orchestrator does persist an error message to the database, but `internal/service/local_service.go` currently drops the `Error` field when building `DownloadStatus`, so the message never actually reaches the frontend. It might be worth forwarding it through the service layer as a follow-up.

anudeepd:
Implemented in PR #595. 

Error messages and metadata now persist through the event/service/CLI layers, and TUI startup restores errored downloads instead of showing completed at 0%. Full Go suite and changed-package race checks pass.

SuperCoolPencil:
Closed by #596

PR #596, opened 2026-08-11
fix: persist and display download errors through restart

superGekFordJ:
Hey there,

Following up on the ideas discussed in issue #591, it seems the bug stems from a few gaps in the error persistence and display chain: the `Error` field isn't persisted on `EventError`/`EventComplete` finalization failures, it's dropped during `GetStatus`/`List` reads, and `dm.err` isn't populated in the TUI during startup restoration.

I want to thank PR #595 for providing some great inspiration on how to tackle this. However, its simplified approach to `EventError` persistence unfortunately introduces some regressions to the core download logic. Specifically, it doesn't preserve chunk-level resume metadata via `SaveStateWithOptions` (meaning errored downloads can't be resumed), and its field-merge logic doesn't distinguish task-backed snapshots, which can overwrite legitimate `Downloaded=0` states.

So, building on the initial issue discussion and the inspiration from #595, I've put together a minimal patch here. It correctly persists the `Error` string on `EventError` (including handling the `existing == nil` case) and finalization failures, propagates it through `GetStatus`/`List` to the TUI, sets `dm.err` during startup restoration, and handles the `cp.GetError()` TOCTOU race. The store layer (`store/state.go`) and the delicate resume-snapshot logic (`SaveStateWithOptions`) are left completely untouched—this patch simply adds the `Error` field to the existing `AddToMasterList` calls, intentionally leaving the heavy lifting for backend state merges to PR #580.

This makes the fix completely independent but highly complementary to #580. Let's use this minimal patch to close the gap on the display side.

*(P.S. If this gets merged first, resolving the git conflict for PR #580 will be straightforward. The fallback branch in #580's `EventError` handler can just absorb the nil-handling and `Error` assignments introduced here. I'll gladly take care of rebasing #580 once this is in!)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Download failures now preserve their underlying error messages and relevant details.
  * Failed downloads are recorded automatically when needed.
  * Status, history, and terminal views now accurately show failed downloads and their error messages.
  * A clear fallback message is displayed when no specific error is available.
* **Tests**
  * Added coverage for error persistence, status reporting, history, and terminal interface behavior.

### DIFF
--- a/internal/orchestrator/events.go
+++ b/internal/orchestrator/events.go
@@ -241,6 +241,7 @@ func (mgr *LifecycleManager) StartEventWorker(ch <-chan types.DownloadEvent) {
 					URLHash:      urlHash,
 					DestPath:     destPath,
 					Filename:     filename,
+					Error:        err.Error(),
 					Status:       "error",
 					TotalSize:    m.Total,
 					Downloaded:   m.Total,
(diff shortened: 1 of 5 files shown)


SuperCoolPencil (approved):
Thanks as always :D

SuperCoolPencil (approved):
The code changes look excellent. They properly persist error messages to the database and correctly expose them to the UI, along with good test coverage handling both legacy and new behavior. I have also verified that the tests pass and there are no linting issues.

PR #600, opened 2026-08-11
Systemd service option

Canikks:
Updated the version number to reflect the latest and added the surge systemd service integration as a nix option.

## Summary by CodeRabbit

- **New Features**
  - Added an option to enable Surge’s systemd service independently.
  - The service starts after network availability, runs Surge in service mode, and automatically restarts after failures.
  - Surge is installed automatically when enabled through the NixOS module.

- **Chores**
  - Updated the flake release version to 0.11.2.
  - Modernized configuration formatting without changing package or overlay behavior.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -54,7 +54,7 @@
           wants = ["network-online.target"];
 
           serviceConfig = {
-            ExecStart = "${pkgs.surge}/bin/surge service __run";
+            ExecStart = "${pkgs.surge}/bin/surge server start --is-system-service";
 
             Restart = "on-failure";
             RestartSec = "5s";
--- a/flake.nix
+++ b/flake.nix
@@ -3,31 +3,64 @@
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
-  outputs = { self, nixpkgs }:
-    let
-      version = "0.8.5";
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -3,31 +3,64 @@
...
+        environment.systemPackages = [pkgs.surge];
+        systemd.services.surge = lib.mkIf cfg.systemd.enable {
+          description = "Surge Download Manager";
+
+          wantedBy = ["multi-user.target"];
+          after = ["network-online.target"];
+          wants = ["network-online.target"];
+
+          serviceConfig = {
+            ExecStart = "${pkgs.surge}/bin/surge service __run";
```
SuperCoolPencil: The NixOS unit should run the daemon directly instead of entering the imperative service-manager wrapper. `surge service __run` calls the kardianos `Service.Run()` path; when the daemon returns an error, that wrapper invokes `systemctl stop surge`. Because systemd treats this as an explicit stop operation, `Restart=on-failure` is not applied, so a transient failure can leave the service stopped despite the restart policy.

Suggested change:

```suggestion
            ExecStart = "${pkgs.surge}/bin/surge server start --is-system-service";
```

The hidden flag preserves Surge’s system-service token/runtime behavior. The `surge service install/start/stop` interface should remain the imperative cross-platform workflow, while the NixOS module owns and runs its unit declaratively.

SuperCoolPencil (approved):

PR #614, opened 2026-08-17
fix: restore the star history chart in the README

Dessalines39394:
The star history chart in the README has been broken by GitHub stargazer API restrictions. This change points it to star-history.dera.page, which provides the same chart backed by an alternative data source that requires no API token, so the chart works again.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README’s Star History chart links to use the new chart hosting domain.


SuperCoolPencil (approved):
Thanks for contributing :)

PR #632, opened 2026-08-24
feat: make concurrent downloads throttle-aware

SuperCoolPencil:
## Summary

- remove duplicate range-request hedging and rely on safe task splitting
- add opt-in adaptive concurrency that backs off on 429/503 responses and recovers gradually
- preserve rate-limited downloads for later recovery while avoiding resumed-download prewarming
- harden range coordination, pause-state delivery, probe locking, and shared transport limits
- reduce the default per-download connection count from 32 to 16
- add deterministic throttle benchmarks and focused regression coverage

## Validation

- `go test ./...`
- `go test -race ./internal/strategy/concurrent ./internal/scheduler ./internal/progress ./internal/transport`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added optional adaptive download concurrency that reduces activity during throttling and gradually recovers.
  * Added settings for adaptive concurrency intervals and dial hedging.
  * Improved pause and resume event handling and state recovery.
  * Improved connection pooling and runtime user-agent handling.

* **Bug Fixes**
  * Improved rate-limit cooldown handling and reduced rapid retries.
  * Prevented incomplete ranged responses from delaying task completion.

* **Documentation**
  * Updated default connection limits and added benchmark instructions.


SuperCoolPencil:
@superGekFordJ applied some of things we talked about a long time ago. Would love if you could take a look at this PR :)

superGekFordJ:
Haha @SuperCoolPencil, I was actually just thinking about bringing the full VerifiedProgress mechanism over to finally close out #579, and it's super cool to see these ideas we've explored before landing in Surge!

I actually feel that AIMD and request hedging can totally coexist here. In GoAria, there were moments I really loved—the download was lagging at like 6–10 MB/s near the very end, and then hedge workers kicked in, instantly shooting the speed up to ~70 MB/s and finishing it right off. So it really feels worth keeping around!

The real issue with hedge before was just the race conditions and flaky state machine under concurrency, which is honestly what the VP mechanism fixes.

To me, both mechanisms have their own sweet spots:
* **Adaptive Tail Splitting** helps split things up and keep idle workers busy when there’s still unsplit data left.
* **Request Hedging** rescues those annoying tail-end stragglers when the last chunk is already in-flight on a slow connection and can't be split anymore.

I'd love to try bringing the VP mechanism into this PR so we can have both working nicely together. Excited to see these two features team up to make downloads even faster for every Surge user.

SuperCoolPencil:
I'll take a look at GoAria and how the VP is implemented. I'll try to see if I can port it over in this PR itself

superGekFordJ:
Don't worry about digging through all the GoAria specifics—I'm actually already halfway through putting together a clean branch based on your PR #632 with the core VP bits & guards. 

I'll push it to a branch / open a sub-PR against your branch shortly so you can just review the clean diff directly.

SuperCoolPencil:
I tried to look into a lot of different ways we could do hedging but it falls into one of 3 problems:
- It's not really hedging, just _smarter_ stealing
- It's gonna create maintainability problems in the future and it's gonna be riddled with data races
- It's gonna make the normal workers be more complex and have more overhead. 
Right now, I'm out of ideas. Will look into it in the future. looks like it's time to merge this PR :)

superGekFordJ:
Yeah, that makes a lot of sense. Merging this PR now is definitely the most practical move.

I can open a separate PR later for the VerifiedProgress mechanism. With VP handling the state and progress safety under the hood, adding this "bonus stealing" on top might actually be pretty straightforward and take very little effort.

Looking forward to seeing this merged!

PR #647, opened 2026-09-08
Add scripts/install.sh for convenient no-package-manager installs

### LINKED ISSUE #646: Add install.sh for convenient Linux/ARM64/musl setup (no brew/AUR/Nix/Docker)
## Problem

The install matrix in `docs/getting-started.md` / the project site only covers users who already have Homebrew, AUR (`yay`), winget/scoop/choco, or Nix. Everyone else — including any generic Linux user, and specifically Alpine-based distros like postmarketOS on ARM64 — is told to just "[Download a release](https://github.com/SurgeDM/Surge/releases/latest)" with no further guidance.

In practice that means manually:
1. Figuring out the right `GOOS`/`GOARCH` naming for your platform
2. Downloading the matching tarball by hand
3. Downloading `Surge_<version>_checksums.txt` and verifying it yourself
4. Extracting and `chmod +x`-ing the binary
5. Moving it somewhere on `PATH`
6. Separately running `surge completion <shell>` and wiring it into your shell config

This is annoying busywork for something the release pipeline (goreleaser) already builds and checksums correctly — including `linux_arm64`, which works great on musl/Alpine (it's a static Go binary, `CGO_ENABLED=0`), it's just undiscoverable from the docs.

## Proposed fix

A small POSIX `scripts/install.sh` that:
- detects OS (linux/darwin) and arch (amd64/arm64/386) via `uname`
- resolves the latest release tag (or `$SURGE_VERSION` for pinning) via the GitHub API
- downloads the matching `Surge_<version>_<os>_<arch>.tar.gz`
- verifies its SHA-256 against the published `_checksums.txt`
- installs the binary to `~/.local/bin` (overridable via `$SURGE_INSTALL_DIR`)
- warns if that directory isn't on `PATH`
- best-effort installs shell completion for zsh/bash/fish based on `$SHELL`

Usage would be:
```sh
curl -sSL https://raw.githubusercontent.com/SurgeDM/Surge/main/scripts/install.sh | sh
```

which mirrors what rustup/homebrew-less installers for lots of other Go CLIs already do, and would slot into the `getting-started.md` table as an "Any Linux/macOS" row.

I wrote and tested this end-to-end on postmarketOS (Alpine/musl) aarch64 — verified it downloads the correct `linux_arm64` asset, the checksum matches, the binary runs, and completion installs correctly. Script attached below; happy to open a PR with it (plus the docs table update) if that's welcome.

```sh
#!/bin/sh
# Surge installer: detects OS/arch, downloads the matching release asset,
# verifies its checksum, and installs the binary (plus shell completion).
#
# Usage:
#   curl -sSL https://raw.githubusercontent.com/SurgeDM/Surge/main/scripts/install.sh | sh
#
# Env overrides:
#   SURGE_INSTALL_DIR   install location for the binary (default: ~/.local/bin, falls back to /usr/local/bin with sudo)
#   SURGE_VERSION        install a specific tag instead of latest, e.g. v0.12.1

set -eu

REPO="SurgeDM/Surge"

log() { printf '%s\n' "$*" >&2; }
die() { log "error: $*"; exit 1; }

need() { command -v "$1" >/dev/null 2>&1 || die "missing required command: $1"; }

need curl
need tar
need sha256sum || need shasum

detect_os() {
  case "$(uname -s)" in
    Linux) echo linux ;;
    Darwin) echo darwin ;;
    *) die "unsupported OS: $(uname -s) (use the Windows installers, or download a release manually)" ;;
  esac
}

detect_arch() {
  case "$(uname -m)" in
    x86_64|amd64) echo amd64 ;;
    aarch64|arm64) echo arm64 ;;
    i386|i686) echo 386 ;;
    *) die "unsupported architecture: $(uname -m)" ;;
  esac
}

OS="$(detect_os)"
ARCH="$(detect_arch)"

VERSION="${SURGE_VERSION:-}"
if [ -z "$VERSION" ]; then
  VERSION="$(curl -sSL "https://api.github.com/repos/${REPO}/releases/latest" \
    | grep '"tag_name":' | head -1 | sed -E 's/.*"([^"]+)".*/\1/')"
  [ -n "$VERSION" ] || die "could not determine latest version"
fi
VERSION_NUM="${VERSION#v}"

ASSET="Surge_${VERSION_NUM}_${OS}_${ARCH}.tar.gz"
BASE_URL="https://github.com/${REPO}/releases/download/${VERSION}"

WORKDIR="$(mktemp -d)"
trap 'rm -rf "$WORKDIR"' EXIT

log "Downloading ${ASSET} (${VERSION})..."
curl -sSL -o "${WORKDIR}/${ASSET}" "${BASE_URL}/${ASSET}" \
  || die "failed to download ${BASE_URL}/${ASSET} (this OS/arch combination may not be published — see https://github.com/${REPO}/releases)"

log "Verifying checksum..."
curl -sSL -o "${WORKDIR}/checksums.txt" "${BASE_URL}/Surge_${VERSION_NUM}_checksums.txt"
(
  cd "$WORKDIR"
  EXPECTED="$(grep " ${ASSET}\$" checksums.txt | awk '{print $1}')"
  [ -n "$EXPECTED" ] || die "no checksum entry for ${ASSET}"
  if command -v sha256sum >/dev/null 2>&1; then
    ACTUAL="$(sha256sum "${ASSET}" | awk '{print $1}')"
  else
    ACTUAL="$(shasum -a 256 "${ASSET}" | awk '{print $1}')"
  fi
  [ "$EXPECTED" = "$ACTUAL" ] || die "checksum mismatch for ${ASSET} (expected ${EXPECTED}, got ${ACTUAL})"
)

tar -xzf "${WORKDIR}/${ASSET}" -C "$WORKDIR"

INSTALL_DIR="${SURGE_INSTALL_DIR:-$HOME/.local/bin}"
mkdir -p "$INSTALL_DIR"
install -m 755 "${WORKDIR}/surge" "${INSTALL_DIR}/surge"
log "Installed surge to ${INSTALL_DIR}/surge"

case ":$PATH:" in
  *":${INSTALL_DIR}:"*) ;;
  *) log "note: ${INSTALL_DIR} is not on your PATH. Add it, e.g.: export PATH=\"${INSTALL_DIR}:\$PATH\"" ;;
esac

# Best-effort shell completion install (non-fatal if it fails).
install_completion() {
  shell="$1"
  case "$shell" in
    zsh)
      dir="${ZDOTDIR:-$HOME}/.zsh/completions"
      mkdir -p "$dir" 2>/dev/null && "${INSTALL_DIR}/surge" completion zsh > "${dir}/_surge" 2>/dev/null \
        && log "Installed zsh completion to ${dir}/_surge (add \"fpath=(${dir} \$fpath)\" before compinit in your .zshrc if not already present)"
      ;;
    bash)
      dir="$HOME/.local/share/bash-completion/completions"
      mkdir -p "$dir" 2>/dev/null && "${INSTALL_DIR}/surge" completion bash > "${dir}/surge" 2>/dev/null \
        && log "Installed bash completion to ${dir}/surge"
      ;;
    fish)
      dir="$HOME/.config/fish/completions"
      mkdir -p "$dir" 2>/dev/null && "${INSTALL_DIR}/surge" completion fish > "${dir}/surge.fish" 2>/dev/null \
        && log "Installed fish completion to ${dir}/surge.fish"
      ;;
  esac
}

case "${SHELL:-}" in
  */zsh) install_completion zsh ;;
  */bash) install_completion bash ;;
  */fish) install_completion fish ;;
esac

log "Done. Run 'surge --version' to confirm."
```

hanthor:
Closes #646.

## Problem

The install matrix in `docs/getting-started.md` only covers brew/AUR/winget/scoop/choco/Nix. Anyone without one of those — including generic Linux and Alpine-based distros like postmarketOS on ARM64 — is told to just "download a release" and figure out the rest by hand: pick the right `GOOS`/`GOARCH` asset name, download the checksums file, verify it, extract, `chmod +x`, place on `PATH`, and separately run `surge completion <shell>`.

## What this adds

`scripts/install.sh`, a POSIX shell script that:

- Detects OS (linux/darwin) and arch (amd64/arm64/386) via `uname`
- Resolves the latest release tag via the GitHub API, or installs a pinned version via `SURGE_VERSION`
- Downloads the matching `Surge_<version>_<os>_<arch>.tar.gz`
- Verifies its SHA-256 against the published `_checksums.txt`
- Installs the binary to `~/.local/bin` (override with `SURGE_INSTALL_DIR`), warning if that directory isn't on `PATH`
- Best-effort installs shell completion for zsh/bash/fish based on `$SHELL`

Usage:

```sh
curl -sSL https://raw.githubusercontent.com/SurgeDM/Surge/main/scripts/install.sh | sh
```

Also updates `docs/getting-started.md` with a row for this and a short explanation.

## Testing

- `shellcheck scripts/install.sh` passes clean.
- Ran the script end-to-end on postmarketOS (Alpine/musl) aarch64: it downloaded `Surge_0.12.1_linux_arm64.tar.gz`, verified the checksum, installed a working binary, and installed zsh completion.
- No Go code changed, so `go test ./...` is unaffected.

## Summary by CodeRabbit

- **New Features**
  - Added a no-package-manager installation option using a shell script.
  - Automatically detects supported operating systems and architectures.
  - Verifies downloaded releases with SHA-256 checksums.
  - Supports custom installation locations and optional version selection.
  - Provides best-effort shell-completion setup for Bash, Zsh, and Fish.

- **Documentation**
  - Updated the getting-started guide with installation, customization, verification, and shell-completion instructions.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -13,7 +13,13 @@ Use the package manager or release artifact that suits your platform:
 | macOS / Linux | `brew install SurgeDM/tap/surge` |
 | Arch Linux | `yay -S surge` |
 | Nix / NixOS | `nix run github:SurgeDM/Surge` |
+| Linux / macOS (no package manager) | `curl -sSL https://raw.githubusercontent.com/SurgeDM/Surge/main/scripts/install.sh \| sh` |
 | Any supported platform | [Download a release](https://github.com/SurgeDM/Surge/releases/latest) |
+
+The install script detects your OS and architecture (including Linux ARM64,
+e.g. Alpine/postmarketOS), downloads the matching release asset, verifies its
+checksum, installs the binary to `~/.local/bin` (override with
+`SURGE_INSTALL_DIR`), and sets up shell completion for zsh/bash/fish.
 
 Run `surge --version` after installing to confirm that your shell can find the
 binary.
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,117 @@
+#!/bin/sh
+# Surge installer: detects OS/arch, downloads the matching release asset,
+# verifies its checksum, and installs the binary (plus shell completion).
+#
+# Usage:
+#   curl -sSL https://raw.githubusercontent.com/SurgeDM/Surge/main/scripts/install.sh | sh
+#
+# Env overrides:
+#   SURGE_INSTALL_DIR   install location for the binary (default: ~/.local/bin, falls back to /usr/local/bin with sudo)
+#   SURGE_VERSION        install a specific tag instead of latest, e.g. v0.12.1
+
+set -eu
+
+REPO="SurgeDM/Surge"
+
+log() { printf '%s\n' "$*" >&2; }
+die() { log "error: $*"; exit 1; }
+
+need() { command -v "$1" >/dev/null 2>&1 || die "missing required command: $1"; }
+
+need curl
+need tar
+need sha256sum || need shasum
+
+detect_os() {
+  case "$(uname -s)" in
+    Linux) echo linux ;;
+    Darwin) echo darwin ;;
+    *) die "unsupported OS: $(uname -s) (use the Windows installers, or download a release manually)" ;;
+  esac
+}
+
+detect_arch() {
+  case "$(uname -m)" in
+    x86_64|amd64) echo amd64 ;;
+    aarch64|arm64) echo arm64 ;;
+    i386|i686) echo 386 ;;
+    *) die "unsupported architecture: $(uname -m)" ;;
+  esac
+}
+
+OS="$(detect_os)"
+ARCH="$(detect_arch)"
+
+VERSION="${SURGE_VERSION:-}"
+if [ -z "$VERSION" ]; then
+  VERSION="$(curl -sSL "https://api.github.com/repos/${REPO}/releases/latest" \
+    | grep '"tag_name":' | head -1 | sed -E 's/.*"([^"]+)".*/\1/')"
+  [ -n "$VERSION" ] || die "could not determine latest version"
+fi
+VERSION_NUM="${VERSION#v}"
+
+ASSET="Surge_${VERSION_NUM}_${OS}_${ARCH}.tar.gz"
+BASE_URL="https://github.com/${REPO}/releases/download/${VERSION}"
+
+WORKDIR="$(mktemp -d)"
+trap 'rm -rf "$WORKDIR"' EXIT
+
+log "Downloading ${ASSET} (${VERSION})..."
+curl -sSL -o "${WORKDIR}/${ASSET}" "${BASE_URL}/${ASSET}" \
+  || die "failed to download ${BASE_URL}/${ASSET} (this OS/arch combination may not be published — see https://github.com/${REPO}/releases)"
+
+log "Verifying checksum..."
+curl -sSL -o "${WORKDIR}/checksums.txt" "${BASE_URL}/Surge_${VERSION_NUM}_checksums.txt"
+(
+  cd "$WORKDIR"
+  EXPECTED="$(grep " ${ASSET}\$" checksums.txt | awk '{print $1}')"
+  [ -n "$EXPECTED" ] || die "no checksum entry for ${ASSET}"
+  if command -v sha256sum >/dev/null 2>&1; then
+    ACTUAL="$(sha256sum "${ASSET}" | awk '{print $1}')"
+  else
+    ACTUAL="$(shasum -a 256 "${ASSET}" | awk '{print $1}')"
+  fi
+  [ "$EXPECTED" = "$ACTUAL" ] || die "checksum mismatch for ${ASSET} (expected ${EXPECTED}, got ${ACTUAL})"
+)
+
+tar -xzf "${WORKDIR}/${ASSET}" -C "$WORKDIR"
+
+INSTALL_DIR="${SURGE_INSTALL_DIR:-$HOME/.local/bin}"
+mkdir -p "$INSTALL_DIR"
+install -m 755 "${WORKDIR}/surge" "${INSTALL_DIR}/surge"
+log "Installed surge to ${INSTALL_DIR}/surge"
+
+case ":$PATH:" in
+  *":${INSTALL_DIR}:"*) ;;
+  *) log "note: ${INSTALL_DIR} is not on your PATH. Add it, e.g.: export PATH=\"${INSTALL_DIR}:\$PATH\"" ;;
+esac
+
+# Best-effort shell completion install (non-fatal if it fails).
+install_completion() {
+  shell="$1"
+  case "$shell" in
+    zsh)
+      dir="${ZDOTDIR:-$HOME}/.zsh/completions"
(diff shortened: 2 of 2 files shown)


junaid2005p:
Hey @hanthor I looked at the PR here are a few inputs
1. The link for the script should be tied to a specific commit id instead of main I think. The reason for this is if due to some malicious activity code inside main is affected even then a malicious script can not be run. In future if updates to the script are done then the link can be modified again to use the new commit Id. 

2. In install.sh need() is implemented as 
```need() { command -v "$1" >/dev/null 2>&1 || die "missing required command: $1"; }```
so if a system has no sha256sum then the program exits and never has a chance to look for shasum
```|| shasum``` is never reached

3. In the zsh completion setup, mkdir and the completion command are chained with &&. If either fails, install_completion returns a non-zero status. Since the script uses set -e, this can cause the installer to exit even though shell completion is intended to be best-effort/non-fatal.

other than that the script looks good to me. Thanks for contributing. Let me know if you need help making these changes.

**Comment by hanthor** [comment]
Thanks @junaid2005p and @coderabbitai for the review! Applied all requested fixes:

1. **Pinned installer URL to immutable commit** – both `docs/getting-started.md` and `scripts/install.sh` (Usage header) now use `https://raw.githubusercontent.com/SurgeDM/Surge/<SHA>/scripts/install.sh` instead of `main`. Current HEAD pins to `22f23af` (parent commit) which already contains the hardened script; future script updates should bump the pin again to the new commit. This prevents TOCTOU if `main` is compromised before archive checksum verification (CWE-494).

2. **Fixed `shasum` fallback unreachable** – `need sha256sum || need shasum` called `die` before fallback. Now:
```sh
if ! command -v sha256sum >/dev/null 2>&1 && ! command -v shasum >/dev/null 2>&1; then
  die "missing required command: sha256sum or shasum"
fi
```
and later `if command -v sha256sum … else shasum -a 256` for verification. Systems with only `shasum` (macOS) now install correctly.

3. **Curl fail on HTTP error** – added `-f` (`curl -fsSL`) to all 3 download sites: GitHub API latest tag fetch, release asset, and checksums.txt. 404/5xx now fail fast instead of feeding an HTML error page to `tar`.

4. **Zsh/bash/fish completion non-fatal under `set -e`** – changed chained `mkdir -p … && surge completion … && log …` to:
```sh
if mkdir -p "$dir" 2>/dev/null && "${INSTALL_DIR}/surge" completion zsh > "${dir}/_surge" 2>/dev/null; then
  log "Installed zsh completion …"
fi
```
(`if` guards `set -e`) for all 3 shells, so a failed `mkdir` or completion generation no longer aborts the installer. The installer already logged completion as best-effort.

Verified `shellcheck scripts/install.sh` passes. No Go code changed.

SuperCoolPencil (dismissed):

SuperCoolPencil:
Follow-up: automatic release selection now scans the releases list and accepts only stable core tags (`vX.Y.Z`). GitHub’s `/releases/latest` endpoint selected `ext-v2.1.3`, which has no platform tarball, causing the 404.

The public install command is now:

```sh
curl -fsSL https://surgedm.github.io/install | sh
```

The GitHub Pages bootstrap is pinned to immutable installer revision `10560f56f0ffc38d264722a38ad78ebbe53c4f76`.

SuperCoolPencil (approved):

